PR #1, opened 2016-02-23
Implementing peering, layers, refactoring the existing ETCD helpers and trying to make this more layer centric/friendly

lazypower:


metadata.yaml
```diff
@@ -6,11 +6,14 @@ description: |
...
+  - keystore
   - layer
+series:
+  - trusty
 subordinate: false
 provides:
   db:
     interface: etcd
-peer:
+peers:
```
mbruzek, line 17: Great catch here! I am sure that was hard to find.

reactive/etcd.py
```diff
@@ -13,19 +13,20 @@
...
-from charmhelpers.core.unitdata import kv
 from charmhelpers.core import templating
-from charmhelpers import fetch
 
 from etcd import EtcdHelper
+from subprocess import CalledProcessError
 
 @hook('config-changed')
 def remove_states():
+    # Matt - this is where it's not immutable :)
```
mbruzek, line 23: That is OK I will forget about the other two configuration options (port, management-port) that are not checked in this method. Perhaps those should be constants rather than config anyway.

reactive/etcd.py
```diff
@@ -36,9 +37,47 @@ def remove_configuration_state():
     cluster_data['leader_address'] = unit_get('private-address')
     leader_set(cluster_data)
 
+
+@when('cluster.joining')
+def cluster_update(cluster):
+    '''' Runs on cluster "disturbed" mode. Each unit is declaring their
```
mbruzek, line 43: Right on! I love the band [Disturbed](https://en.wikipedia.org/wiki/Disturbed_%28band%29). Oh perhaps you meant "distributed".

reactive/etcd.py
```diff
@@ -36,9 +37,47 @@ def remove_configuration_state():
...
+            'unit_name': cluster.unit_name()}}
+    # Store data about every peer we have seen. this may be useful later
+    etcd_helper.cluster_data(unit)
+    remove_state('etcd.configured')
+    if is_leader():
+        # store and leader-set the new cluster string
+        leader_set({'cluster': etcd_helper.cluster_string()})
+        remove_state('cluster.joining')
+
+
```
mbruzek, line 58: Too many line returns, linting tools do not approve.  ( flake8- :cop: )

lib/etcd.py
```diff
@@ -60,33 +75,89 @@ def fetch_and_install(self, source, sha):
...
+                log('Notice: Unit failed to registration command', 'WARNING')
+
+    def unregister(self, cluster_data):
+        # this wont work, it needs to be etcdctl member remove {{GUID}}
+        command = "etcdctl -C http://{}:{} member remove {}" \
+                  " http://{}:{}".format(cluster_data['leader_address'],
+                                         config('port'),
+                                         cluster_data['private_address'],
+                                         config('management_port'))
+        check_call(split(command))
```
mbruzek, line 149: If this command does not work it will need a similar, try/except as the register() function. I understand there is more work t be done on scale down, but lets not make it throw an error on the way out.

mbruzek:
There are some minor things I found but none of them should hold up this merge.

reactive/etcd.py
lazypower, line 23: Constants leave something to be desired really, as they should be configurable depending on the colocation needs

lazypower:
Thanks for the review. I added you to the collaborators so you can merge now if its good to you

PR #2, opened 2016-02-23
Added client-string generator

lazypower:
Adds a string builder to aid when handling *-relation-joined/changed
to hand out a cluster connection string understood by both families of
uses.

eg: docker expects etcd:// for swarm discovery
eg: everything else on the planet uses http:// or https:// urls


mbruzek:
+1 LGTM

PR #3, opened 2016-02-24
Added client-string generator

lazypower:
Adds a string builder to aid when handling *-relation-joined/changed to
    hand out a cluster connection string understood by both families of uses.

```
eg: docker expects etcd:// for swarm discovery eg: everything else on the
planet uses http:// or https:// urls
```
-  Adds icon, new interface declaration etcd-cluster
-  Import the legacy items from the original charm

### DIFF
--- a/actions/health
+++ b/actions/health
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -e
+PRIVATE_ADDRESS=$(unit-get private-address)
+OUT=$(/opt/etcd/etcdctl -C http://${PRIVATE_ADDRESS}:4001 cluster-health)
+
+action-set result-map.message="${OUT}"
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -16,7 +16,7 @@ from charmhelpers.core import host
 from charmhelpers.core import templating
 
 from etcd import EtcdHelper
-from subprocess import CalledProcessError
+
 
 @hook('config-changed')
 def remove_states():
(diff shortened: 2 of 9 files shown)


lazypower:
Related to work going on in https://github.com/chuckbutler/interface-etcd/pull/1

actions/health
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+PRIVATE_ADDRESS=$(unit-get private-address)
+OUT=$(/opt/etcd/etcdctl -C http://${PRIVATE_ADDRESS}:4001 cluster-health)
```
mbruzek, line 5: fancy!

reactive/etcd.py
```diff
@@ -68,6 +66,16 @@ def remove_unit_from_cluster(cluster):
     remove_state('cluster.departed')
 
 
+@when('db.connected')
```
mbruzek, line 69: Who sets this state?

mbruzek, line 69: interface is updated to set this state

tests/10-deploy
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_two_node_scale(self):
+        self.deployment.add_unit('etcd')
+        self.deployment.sentry.wait()
+
+        status1 = self.deployment.sentry['etcd'][0].run('service etcd status')
+        status2 = self.deployment.sentry['etcd'][1].run('service etcd status')
+        self.assertTrue("running" in status1[0])
+        self.assertTrue("running" in status2[0])
+
```
mbruzek, line 33: running != healthy cluster.  You could also run a `etcdctl cluster-health` on one of the machines here to make sure the cluster is healthy. Improvements for next iteration.

mbruzek:
I have not tested this latest iteration but +1 LGTM

PR #6, opened 2016-02-29
Remove etcd cluster

lazypower:
Removing an interface that needs to be killed with :fire:

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -38,21 +38,30 @@ def remove_configuration_state():
     leader_set(cluster_data)
 
 
+@when('cluster.prepare_to_launch')
+def cluster_declaration(cluster):
+    etcd = EtcdHelper()
+    for unit in cluster.list_peers():
+        cluster.provide_cluster_details(scope=unit,
+                                        public_address=etcd.public_address,
+                                        port=etcd.port,
+                                        unit_name=etcd.unit_name)
+
+
 @when('cluster.joining')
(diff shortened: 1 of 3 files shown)


reactive/etcd.py
```diff
@@ -38,22 +38,34 @@ def remove_configuration_state():
...
+                        'unit_name': cluster.unit_name()}})
+
     if is_leader():
         # store and leader-set the new cluster string
-        leader_set({'cluster': etcd_helper.cluster_string()})
-        remove_state('cluster.joining')
+        leader_set({'cluster': etcd.cluster_string()})
+
+    remove_state('etcd.configured')
+    remove_state('cluster.joining')
```
mbruzek, line 68: moving this out one level seems to be the right thing.

reactive/etcd.py
```diff
@@ -38,22 +38,34 @@ def remove_configuration_state():
...
+@when('cluster.declare_self')
+def cluster_declaration(cluster):
+    etcd = EtcdHelper()
+    for unit in cluster.list_peers():
+        cluster.provide_cluster_details(scope=unit,
+                                        public_address=etcd.public_address,
+                                        port=etcd.port,
+                                        unit_name=etcd.unit_name)
+
+    remove_state('cluster.declare_self')
```
mbruzek, line 50: Makes sense

mbruzek:
+1 LGTM

PR #9, opened 2016-05-03
Fix failure due to recent path.py upgrade

lazypower:
Path.py as it continues progressing forward into intrepid territory, has moved on to a newer setuptools. I dont have a good idea of if we should pin this, so I'm going to run with the assumption of go with stable until the build breaks, then pin it.


marcoceppi:
LGTM

lazypower:
PASS: 4 Total: 4 (765.893144 sec)

mbruzek:
+1

mbruzek:
Remember to rev the bundles!  (self)

Issue #8, opened 2016-04-18
Import Xenial changes from project-calico


chuckbutler:
https://code.launchpad.net/~project-calico/charms/xenial/etcd/trunk

chuckbutler:
Xenial support has landed, closing

PR #10, opened 2016-05-10
Xenial / Systemd / Offline support

lazypower:
This branch introduces some very different behavior from the install mechanism, but overall simplifies delivery of the charm.
- Offlineable resources via juju_resources
- Support for SystemD based hosts (verified on Xenial)
- Cross arch support via juju_resources

Stems from #8


README.md
```diff
@@ -23,7 +23,51 @@ juju add-unit -n 2 etcd
...
+## Usage with your own binaries
+
+This charm supports resources, which means you can supply your own release of
+etcd and etcdctl to this charm. Which by nature makes it highly deployable on
+multiple architectures.
+
+### Supply your own binaries
+
+```shell
+juju upgrade-charm etcd --resource etcd=./path/to/etcd --reosurce etcdtcl=./path/to/etcdctl
```
nelljerram: Typo 'reosurce'

README.md
```diff
@@ -23,7 +23,51 @@ juju add-unit -n 2 etcd
...
+juju list-resources etcd
+```
+
+You will see your binaries have been provided by username@local, and the charm
+will reconfigure itself to deploy the provided binaries.
+
+### If you do not have the binaries
+
+If you are simply deploying etcd, and the charm has halted demanding resources by
+telling you to consult the READJE, you can use a script contained in the charm
```
nelljerram: Typo 'READJE'

README.md
```diff
@@ -23,7 +23,51 @@ juju add-unit -n 2 etcd
 
 Its recommended to run an odd number of machines as it has greater redundancy than even number (ie. 4, you can lose 1 before quorum is lost, where as 5, you can 2).
 
+# The charm told me to see the README
 
+## Usage with your own binaries
+
+This charm supports resources, which means you can supply your own release of
+etcd and etcdctl to this charm. Which by nature makes it highly deployable on
+multiple architectures.
```
nelljerram, line 49: As also would just referencing a package name, surely?

README.md
```diff
@@ -23,7 +23,51 @@ juju add-unit -n 2 etcd
 
 Its recommended to run an odd number of machines as it has greater redundancy than even number (ie. 4, you can lose 1 before quorum is lost, where as 5, you can 2).
 
+# The charm told me to see the README
 
+## Usage with your own binaries
+
+This charm supports resources, which means you can supply your own release of
```
nelljerram, line 47: 'can' or 'have to'?  If it's optional, why would anyone ever want to do that, and why would the charm allow them to?

README.md
```diff
@@ -23,7 +23,51 @@ juju add-unit -n 2 etcd
 
 Its recommended to run an odd number of machines as it has greater redundancy than even number (ie. 4, you can lose 1 before quorum is lost, where as 5, you can 2).
 
+# The charm told me to see the README
```
nelljerram, line 38: So here's what I hope the situation is, after all these changes.
- If you're deploying on Xenial or later, the charm installs the etcd package from the Ubuntu archive, for any supported architecture.
- Otherwise, if you're deploying on amd64, the charm grabs and uses a binary tarball from the CoreOS github account.
- Otherwise - i.e. both old distro and weird architecture - you have to provide the resources yourself.

If that's right, here's what I think the README should say:

You've been directed here because you're deploying this etcd charm onto a pre-Xenial and non-amd64-based Ubuntu platform, and we don't have etcd and etcdctl binaries for that platform.  You will need to obtain such binaries somehow yourself (e.g. by downloading and building from source), then tell Juju about those binaries as detailed below.

README.md
lazypower, line 49: No. Juju resources are decoupled from package management.  This provides a streamlined way to provide upstream binaries to the charm. It has a fallback mechanism on xenial hosts to reference archive if no resources are supplied.

This wont work on trusty, as there is no etcd package in archive to support the request. This created the install tree that has 3 paths.

Upstart + resources
Systemd + resources
Systemd + archive packages

Its not implicitly called out this way, as its preferrable to use the resources that will ship with the charm. As a charm maintainer, we can rev the resources on a development channel of the charm and run smoke tests to verify everything is good, and then upgrade the resource payload on the stable channel of the charm to deliver the latest resource. 

This enables offline deployments as well as supporting upstreams best efforts to have patch releases pushed quickly to consumers of the charm.

README.md
lazypower, line 47: 'can'.  Its optional as you may be running on PPC64EL architecture and need to supply a non AMD64 bin for the workload to function, or perhaps the charm store is still deploying v 2.3.3 and v3.0.0 just hit GA and you wish to smoke test it in your own environment. This prevents consumers from having to rev packaging/the-charm to do these operations.

Makefile
```diff
@@ -0,0 +1,28 @@
...
+	@PATH=.tox/py34/bin:.tox/py35/bin flake8 $(wildcard hooks reactive lib unit_tests tests)
+	@charm proof
+
+.PHONY: unit_test
+unit_test: apt_prereqs
+	@echo Starting tests...
+	tox
+
+.PHONY: fetch_resources
+fetch_resources:
```
nelljerram, line 27: I don't get why this is a Makefile target.  IIUC correctly, the charm itself should automatically determine if it needs to download resources, and if so, do that at the appropriate time.  It shouldn't be presented as a user choice.

scripts/download_resources.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+#!/bin/bash
```
nelljerram: Second hashbang line here.

lazypower:
I think my README blobs were a bit nebulous about what resources are actually providing. Here's a link to the release notes version of the feature which hopefully helps explain in better detail while we sort out the verbiage:

https://jujucharms.com/docs/devel/temp-release-notes#resources-a-new-concept-has-been-introduced-into-charms-called

README.md
lazypower, line 38: That sounds far better than how I was trying to articulate it. Thanks :)

reactive/etcd.py
```diff
@@ -180,10 +180,12 @@ def configure_etcd():
                           cluster_data, owner='root', group='root')
     else:
         # render systemd
-        templating.render('systemd', 'files/systemd/etcd.service',
+        templating.render('systemd', '/etc/systemd/system/etcd.service',
                           cluster_data, owner='root', group='root')  # noqa
```
nelljerram: This looks wrong to me, because on Xenial there should already be an etcd.service file installed by the package (in /usr/lib/systemd/system), and the 'systemctl enable etcd' call just below will symlink that into /etc/systemd/system.  So the charm shouldn't need to generate an independent copy of the service file.

templates/defaults
```diff
@@ -4,7 +4,7 @@ ETCD_ADVERTISE_CLIENT_URLS="http://{{ private_address }}:{{ port }}"
 ETCD_LISTEN_CLIENT_URLS="http://127.0.0.1:{{ port }},http://{{ private_address }}:{{ port }}"
 ETCD_LISTEN_PEER_URLS="http://{{ private_address }}:{{ management_port }}"
 ETCD_INITIAL_ADVERTISE_PEER_URLS="http://{{ private_address }}:{{ management_port }}"
-ETCD_ADVERTISE_PEER_URLS="http://172.31.20.186:2380"
+ETCD_ADVERTISE_PEER_URLS="http://{{ private_address}}:{{ management_port }}"
```
nelljerram: Does it matter that there isn't a trailing space after private_address?

README.md
```diff
@@ -1,11 +1,30 @@
 # Etcd
 
-> This charm was built from layers! To see the developer docs see `hacking.md` in
-the charm root.
+Etcd is a highly available distributed key value store that provides a reliable way to store data across a cluster of machines. Etcd gracefully handles master elections during network partitions and will tolerate machine failure, including the master.
+
```
nelljerram, line 7: I personally prefer text wrapped <=79 columns (e.g. to read more easily in gitk).

README.md
```diff
@@ -1,11 +1,30 @@
...
+Etcd allows storing data in a distributed hierarchical database with observation.
+
+# Usage
+
+We can deploy a single node easily with
+
+```shell
+juju deploy etcd
+```
+Add and capacity with:
```
nelljerram: Should be 'And add', I think.

README.md
```diff
@@ -1,11 +1,30 @@
...
+
+```shell
+juju add-unit -n 2 etcd
+```
+
+Its recommended to run an odd number of machines as it has greater redundancy than even number (ie. 4, you can lose 1 before quorum is lost, where as 5, you can 2).
```
nelljerram: Should be It's
s/even/an even/
s/4/with 4/
s/where as 5, you can/whereas with 5, you can lose/

nelljerram:
Thanks for this, @chuckbutler !  Can you advise how I can test it?  Currently I have in my bundle 

```
charm: "cs:~project-calico/xenial/etcd"
```

Can I just change that to:

```
branch: "http://github.com/chuckbutler/layer-etcd;offline-resources"
```

Or do I need some wrapper around layer-etcd?

lazypower:
Thanks for the detailed review @neiljerram. You've surfaced a lot of good points.

I'll give this another go and ping when its ready for re-eval.

lazypower:
Ah and I missed the last comment.

As this is a layer, you will need to build the charm from the layer. If you have charm-tools >= 2.1.2

```
git clone https://github.com/chuckbutler/layer-etcd
cd etcd && git checkout offline-resources
charm build -l DEBUG
```

It will assemble the charm for deployment testing. I can follow up on the review comments and publish to my namespace for easier testing if thats preferable

reactive/etcd.py
lazypower: Ah thats true. We only need to install this file if we've done this via resources. This did cause me some headache and I think I missed a patch I had made remotely to fix this. Good catch

lazypower:
And with that last push, i've revved my namespace of the charm

url: cs:~lazypower/etcd-6

It has the two resources attached: etcd-0  and etcdctl-0 - both of which are from CoreOS etcd 2.3.3 amd6 4 linux release

 juju deploy cs:~lazypower/etcd --series=xenial
or
juju deploy cs:~lazypower/etcd --series=trusty

Both of which will work OOTB with the resources provided. If you wish to test the trusty fallback:

charm pull cs:~lazypower/etcd-6

then juju deploy this version without supplying any of the resources with --series=xenial

Makefile
lazypower, line 27: This is more for my sake, while I'm revving the charm testing different bins and preparing for a publish this script keeps me honest. I've also leveraged it in the README for users wishing to explore the functionality. (it wasn't originally worded this way)

its safe to ignore from a user perspective, but important from a release management perspective.

lazypower:
I just ran into https://bugs.launchpad.net/juju-core/+bug/1577415 while testing the commands. This appears to be an issue with the agent attempting to pull resources from the store, and its a known issue. Will track and gate the publishing of this revision pending a resolution of that bug.

But this still works for locally supplied/denied resources for the charm. so i think there's still a good opportunity for this to be resolved in time.

reactive/etcd.py
```diff
@@ -97,13 +105,53 @@ def update_cluster_string():
...
+        install(etcd_path, '/usr/bin/etcd')
+        install(etcdctl_path, '/usr/bin/etcdctl')
+
+        host.add_group('etcd')
+
+        if not host.user_exists('etcd'):
+            host.adduser('etcd')
+            host.add_user_to_group('etcd', 'etcd')
+
+        os.makedirs('/var/lib/etcd/', exist_ok=True)
```
mbruzek, line 141: exist_ok is only valid in 3.2 and up, making this a 2.7 incompatible change. I see you changed the shebang to python3 so this is fine, but it takes only one more line to make it compatible with 2.7 (here)

``` python
if not os.isdir('/var/lib/etcd')
    os.makedirs('/var/lib/etcd)
```

mbruzek:
I mistakenly deployed this charm with juju 1.25.5 and got an error on the install hook. I now realize the README says I must use juju 2+ to get the resource feature.

I got the following pastebin:  http://pastebin.ubuntu.com/16362633/

It should be possible to catch "NotImplementedError" on line 112 in install_etcd to print some status  messages for the verison 1 users.

reactive/etcd.py
```diff
@@ -97,13 +105,57 @@ def update_cluster_string():
...
-    etcd_helper.fetch_and_install(source, sha)
-    set_state('etcd.installed')
+
+    codename = host.lsb_release()['DISTRIB_CODENAME']
+
+    try:
+        etcd_path = resource_get('etcd')
+        etcdctl_path = resource_get('etcdctl')
+    except NotImplementedError:
+        status_set('blocked', 'This charm requires the resource feature available in juju 2+')  # noqa
```
mbruzek, line 116: This line will give juju 1.x users a good message to get on the 2 train!

reactive/etcd.py
lazypower, line 141: Thats as fair point, but I'd rather let python 2.7 silently die off. Charms  have migrated to python3 since we started the layer approach. I'd like to keep pushing the envelope of making the transition.

templates/systemd
```diff
@@ -0,0 +1,22 @@
...
+
+[Service]
+Environment=DAEMON_ARGS=
+Environment=ETCD_NAME=%H
+Environment=ETCD_DATA_DIR=/var/lib/etcd/
+EnvironmentFile=-/etc/default/%p
+Type=notify
+User=etcd
+PermissionsStartOnly=true
+ExecStart=/usr/bin/etcd $DAEMON_ARGS
```
mbruzek, line 15: I don't systemd very well yet. Should you not use the Jinja template variables in this file as well? How are the default values getting in this systemd file?

lazypower, line 15: line 11 points it at /etc/default/etcd, which is the rendered form of the defaults template

mbruzek:
Great work here @chuckbutler!  

After deploying this on Juju 2+ on both xenial and trusty I do approve of this change. You did quite a bit of work on this and it is very good.

I have not deployed this with charm-store resources which @chuckbutler tells me is blocked by bug: https://bugs.launchpad.net/juju-core/+bug/1577415

README.md
nelljerram, line 49: OK - thanks for the explanation.

Makefile
nelljerram, line 27: Understood - thanks.

nelljerram:
@chuckbutler Just to check something - in your comment above where you wrote

```
If you wish to test the trusty fallback:
```

I think you meant

```
If you wish to test the xenial fallback:
```

Right?

nelljerram:
FYI, proceeding to test this now myself, using cs:~lazypower/etcd-6 ...

lazypower:
@neiljerram thats correct.

Ensure you provide the binaries, or clone the charm if you're looking to test the Xenial archive fallback. The documented bug was discovered trying to use the resources provided by the store. The feature is so new its still shaking out some bugs. pardon our 🚧  construction 👷

PR #12, opened 2016-05-13
WIP - TLS Branch

lazypower:
This branch completes server side, peer to peer, and client communication TLS termination.
- Adds cluster health messaging
- Overrides status_set with "(leader) " for condensed identification
- TLS certificates are now generated with ClientAuth/ServerAuth so server keys may be used to communicate with the cluster
- Updates DEFAULTS template with TLS components

TODO Items remaining: 
- [ ] Will upgrade a running cluster, but will not properly upgrade the cluster string, resulting in an unhealthy cluster
- [x] Does not properly work yet on Trusty
- [ ] Does not hand off client certificates over interface: etcd
- [x] Tests are now horribly broken, and need refactored to expect/work-with the TLS certificates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 resources
+.tox
+.coverage
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 resources
 .tox
 .coverage
+.cache
+__pycache__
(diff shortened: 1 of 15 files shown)


lazypower:
This has diverged, closing in favor of the work condensed and applied on #17

PR #13, opened 2016-05-17
Re-add the etcd-proxy relation

lazypower:
This PR re-introduces the etcd-proxy interface and relationship intended for Etcd to distribute its cluster-string (administrative initial-cluster) to remote proxy units, that will bind to localhost and forward traffic to the etcd cluster.

Read more about this feature here: https://coreos.com/etcd/docs/latest/proxy.html

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -96,6 +96,12 @@ def send_connection_details(client):
     client.provide_connection_string(hosts, config('port'))
 
 
+@when('proxy.connected')
+def send_cluster_details(proxy):
+    etcd = EtcdHelper()
+    proxy.provide_cluster_string(etcd.cluster_string())
+
+
 @hook('leader-settings-changed')
 def update_cluster_string():
     # When the leader makes a broadcast, assume an upset and prepare for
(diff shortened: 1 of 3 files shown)


reactive/etcd.py
```diff
@@ -96,6 +96,12 @@ def send_connection_details(client):
     client.provide_connection_string(hosts, config('port'))
 
 
+@when('proxy.connected')
+def send_cluster_details(proxy):
```
mbruzek, line 100: why you no docstring bro?

mbruzek:
Not trying to be dense here but i don't understand what the difference is between relations, they both seem to send the connection string around. 

Also this change could use a readme update rather  than letting us guess at which etcd relation to use.

lazypower:
One sends the client connection string, and the proxy sends the administrative cluster string. 

its a single word difference in the code and is a hair confusing, even to myself as the author. When initially porting the code over I omitted this as I thought we could use the db relationship for it all. But thats not the case. It needs the resulting --initial-cluster-string in order to discover all the nodes and set itself up as a proxy.

PR #14, opened 2016-05-24
Backport the fix for xenial package cluster initialization

lazypower:
This backports the bits which remove the single-node initialization (this only affects first deploy, from the xenial package).

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -3,6 +3,7 @@
 from charms.reactive import when
 from charms.reactive import when_not
 from charms.reactive import set_state
+from charms.reactive import is_state
 from charms.reactive import remove_state
 from charms.reactive import hook
 
@@ -24,6 +25,7 @@ from pwd import getpwnam
 from subprocess import check_call
 from subprocess import CalledProcessError
 from shlex import split
+from shutil import rmtree
 
 import os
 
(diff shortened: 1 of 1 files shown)


reactive/etcd.py
```diff
@@ -129,6 +131,14 @@ def install_etcd():
...
+            # Stop the service and remove the defaults
+            # I hate that I have to do this. Sorry short-lived local data #RIP
+            # State control is to prevent upgrade-charm from nuking cluster
+            # data.
+            if not is_state('etcd.package.adjusted'):
+                host.service('stop', 'etcd')
+                rmtree('/var/lib/etcd/default')
+                set_state('etcd.package.adjusted')
             set_state('etcd.installed')
             return
```
mbruzek, line 143: Don't you need to start the etcd service before returning from the "install_etcd" method here? The service was running, I don't know what would start it again.

mbruzek:
I hate to delete data like that but I know of no other way to fix it.

reactive/etcd.py
lazypower, line 143: nah, here's why

https://github.com/chuckbutler/layer-etcd/blob/313626c724318c444ad7d71904f7aa22aec72a94/reactive/etcd.py#L216

mbruzek:
I knew the restart was somewhere. I just could not find it. +1

lazypower:
Yep i totally get that. If we re-start before we have inserted our configuration data for the static-cluster style of turn-up, we wind up in the exact same boat as we started in using the xenial package which is targted at a single-node localhost style deployment.

Glad you were keeping me honest sir 👍

PR #15, opened 2016-05-24
Moves xenial to the top of the list

lazypower:
Juju seems to favor the first listed series in metadata unless you
explicitly override it on the cli/bundle. This change defaults to using
xenial which has a clean install path if no resources are available.


mbruzek:
+1 Xenial all the things!

Issue #16, opened 2016-06-02
Quorem Process Broken among peers as of rev-2


chuckbutler:
There has been several reports of the ETCD Quroem failing during turn up of multiple units at once. 

Opening a centralized bug here as this will touch multiple portions of the code base between the reactive layer, interface data, and supporting bundles.

chuckbutler:
Quorem states are looking good in the master branch. This is pending a release, so I will leave this open - however developers deploying from the layer in master will have a more predictable clustering quorem now that we are communicating with the leader directly, and closely executing the cluster-management commands on the leader on behalf of its peers.

chuckbutler:
Quorem states have been completely re-tooled and now work as expected with both scale-up and scale-down scenarios. 

You can even destroy the leader and watch the cluster re-form and populate, however you will still lose PKI on the tls terminated endpoints which leaves you in an inconsistent state, it is still not recommended to terminate the leader.

PR #17, opened 2016-06-03
Huge Merge

### LINKED ISSUE #5: Cant scale down :(
Etcd supports member de registration. 

`etcdctl member unregister {{hash}}`

lazypower:
This branch will not fully deploy cleanly - as it does not provide client certificates to db clients, nor does it provide client certs for the proxy relation.

This branch is an attempt at cleaning the divergence happening in #12  
- Removes all traces of lib/etcd.py to eliminate stale cache issues
- Adds proper etcdctl wrapper to interface directly with ETCD
- Uses etcd-leader response data to bootstrap new units
- Properly halts if a unit is currently cycling during the registration sequence
- Can scale down fixes #5
- touches #16


reactive/etcd.py
```diff
@@ -1,115 +1,77 @@
...
-    proxy.provide_cluster_string(etcd.cluster_string())
-
-
-@hook('leader-settings-changed')
-def update_cluster_string():
-    # When the leader makes a broadcast, assume an upset and prepare for
-    # service restart
-    remove_state('etcd.configured')
 
+# @when('db.connected')
```
mbruzek, line 60: TODO: remove dead code.

reactive/etcd.py
```diff
@@ -175,58 +145,190 @@ def install_etcd():
...
+    static turnup rules. If we find a GUID in the member list without peering
+    information the unit will enter a race condition and must wait for a clean
+    status output before we can progress to self registration.
+    '''
+    # We're going to communicate with the leader, and we need our bootstrap
+    # startup string once.. TBD after that.
+    etcdctl = EtcdCtl()
+    bag = EtcdDatabag()
+    # Assume a hiccup during registration and attempt a retry
+    if bag.cluster_unit_id:
```
mbruzek, line 165: It is not clear to me how cluster_unit_id is being set or not in the initalizer. It seems to be an important variable in this method. Extra comments here would go far to help me understand how EtcdDatabag could either set or not set cluster_unit_id.
EDIT: I scrolled back up to the etcd_databag.py file to see how this was set in the constructor. I see that it comes from the key value store in the charm. Still think some additional comments would be helpful here so you don't have to open the other file to understand what this is and what it means to be set or unset.

lib/etcd_databag.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # These are hard coded, smell for now.
+        self.ca_certificate = "/etc/ssl/etcd/ca.pem"
+        self.server_certificate = "/etc/ssl/etcd/server.pem"
+        self.server_key = "/etc/ssl/etcd/server-key.pem"
+        # Cluster concerns
+        self.token = self.cluster_token()
+        self.cluster_state = 'existing'
+
+        self.cluster_unit_id = self.db.get('cluster_unit_id')
```
mbruzek, line 47: I see cluster_unit_id is set if the db contains it here. I don't know what the existence of cluster_unit_id means at this point. See my comment in etcd.py.

lib/etcdctl.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class EtcdCtl:
+    ''' etcdctl modeled as a python class. This python wrapper consumes
+    and exposes some of the commands contained in etcdctl. Related to unit
+    registration, cluster health, and other operations '''
+
+    def register(self, cluster_data):
+        ''' Perform self registration against the etcd leader and returns the
+        raw output response.
```
mbruzek, line 15: The method does not return raw output, the docstring is not accurate. Suggest  ... and returns None if registration fails, or a registration dictionary with the parsed cluster_unit_id, and initial cluster string.

reactive/etcd.py
```diff
@@ -175,58 +145,190 @@ def install_etcd():
...
+
+    peers = etcdctl.member_list(leader_get('leader_address'))
+    for unit in peers:
+        if 'client_urls' not in peers[unit].keys():
+            # we cannot register. State not attainable.
+            msg = 'Waiting for unit to complete registration'
+            status_set('waiting', msg)
+            return
+
+    if not bag.cluster_unit_id:
```
mbruzek, line 179: This if is the inverse of the if on line 165. Is it possible to use else here? Or does the check for client_urls have to happen inbetween the two blocks?

reactive/etcd.py
```diff
@@ -175,58 +145,190 @@ def install_etcd():
...
+
+            bag.cluster_unit_id = resp['cluster_unit_id']
+            bag.cluster = resp['cluster']
+
+    render('defaults', '/etc/default/etcd', bag.__dict__)
+    host.service_restart('etcd')
+    time.sleep(2)
+
+    # Check health status before we say we are good
+    etcdctl = EtcdCtl()
```
mbruzek, line 196: Why is it necessary to re-create an EtcdCtl object here? A comment to why a new object is necessary is more helpful than the general comment (which I am totally guilty of myself).

reactive/etcd.py
```diff
@@ -175,58 +145,190 @@ def install_etcd():
...
-        templating.render('defaults', '/etc/default/etcd',
-                          cluster_data, owner='root', group='root')
 
-    host.service('restart', 'etcd')
-    set_state('etcd.configured')
+@when('etcd.installed')
+@when('etcd.ssl.placed')
+@when('leadership.is_leader')
+@when_not('etcd.leader.configured')
+def initialize_new_leader():
```
mbruzek, line 209: A docstring would be nice here as the function is non trivial.

reactive/etcd.py
```diff
@@ -175,58 +145,190 @@ def install_etcd():
...
+    bag = EtcdDatabag()
+    bag.token = bag.token
+    bag.cluster_state = 'new'
+    bag.cluster = "{}=https://{}:{}".format(bag.unit_name,
+                                            bag.private_address,
+                                            bag.management_port)
+    render('defaults', '/etc/default/etcd', bag.__dict__, owner='root',
```
mbruzek, line 216: This is the third time I have encountered the same render, restart, sleep combination in this method. That is always a sign that we could abstract these steps to their own method. Suggest something along the lines of:

``` python
def render_restart_etcd(file, options, seconds):
    """Render the configuration file with the options and sleep for the provided seconds."""
    render('defaults', file, options, owner='root', group='root')
    host.service_restart('etcd')
    time.sleep(2)
```

Obviously if the "defaults" file can change this could also be configurable.

lib/etcdctl.py
```diff
@@ -0,0 +1,106 @@
...
+            if 'Added member' in line:
+                reg['cluster_unit_id'] = line.split('ID')[-1].strip(' ').split(' ')[0]  # noqa
+            if 'ETCD_INITIAL_CLUSTER=' in line:
+                reg['cluster'] = line.split('="')[-1].rstrip('"')
+        return reg
+
+    def unregister(self, unit_id):
+        ''' Perform self deregistration during unit teardown
+
+        @params cluster_data - a dict of data to fill out the request to push
```
mbruzek, line 51: Remove this parameter description as it is no longer used.

mbruzek:
This is a huge code change and overall seems very good +1! I left several minor comments that should not prevent a merge, but considered for future improvements.  Running tests now will merge this if tests come successful.

mbruzek:
I was able to spin up a cluster of 9 without any issues. If we find any bugs we should open them for individual problems.

Issue #18, opened 2016-06-03
Error:  client: etcd cluster is unavailable or misconfigured


mbruzek:
I encountered an edge case where the 11th etcd unit didn't register.

```
unit-etcd-10: 2016-06-03 21:35:39 INFO unit.etcd/10.certificates-relation-changed logger.go:40 context deadline exceeded
unit-etcd-10: 2016-06-03 21:35:39 WARNING unit.etcd/10.juju-log server.go:269 certificates:0: Notice:  Unit failed self registration
unit-etcd-10: 2016-06-03 21:35:39 INFO unit.etcd/10.juju-log server.go:269 certificates:0: Writing file /etc/default/etcd root:root 444
unit-etcd-10: 2016-06-03 21:35:39 INFO unit.etcd/10.certificates-relation-changed logger.go:40 Job for etcd.service failed because the control process exited with error code. See "systemctl status etcd.service" and "journalctl -xe" 
for details.
unit-etcd-10: 2016-06-03 21:35:41 INFO unit.etcd/10.certificates-relation-changed logger.go:40 Error:  client: etcd cluster is unavailable or misconfigured
unit-etcd-10: 2016-06-03 21:35:41 INFO unit.etcd/10.certificates-relation-changed logger.go:40 error #0: dial tcp 127.0.0.1:4001: getsockopt: connection refused
unit-etcd-10: 2016-06-03 21:35:41 INFO unit.etcd/10.certificates-relation-changed logger.go:40 error #1: dial tcp 127.0.0.1:2379: getsockopt: connection refused
unit-etcd-10: 2016-06-03 21:35:41 INFO unit.etcd/10.certificates-relation-changed logger.go:40 
unit-etcd-10: 2016-06-03 21:35:41 INFO unit.etcd/10.certificates-relation-changed logger.go:40 Traceback (most recent call last):
unit-etcd-10: 2016-06-03 21:35:41 INFO unit.etcd/10.certificates-relation-changed logger.go:40   File "/var/lib/juju/agents/unit-etcd-10/charm/hooks/certificates-relation-changed", line 19, in <module>
unit-etcd-10: 2016-06-03 21:35:41 INFO unit.etcd/10.certificates-relation-changed logger.go:40     main()
unit-etcd-10: 2016-06-03 21:35:41 INFO unit.etcd/10.certificates-relation-changed logger.go:40   File "/usr/local/lib/python3.5/dist-packages/charms/reactive/__init__.py", line 73, in main
[...]

PR #19, opened 2016-06-03
Catch callprocesserror

mbruzek:
Addresses issue #18


mbruzek:
I checked all the use of cluster_health() and this should work for all cases. I would like to investigate standardizing the cluster_health() uses and possibly re-write this catch to include real output but I didn't have the time to find all output possibilities.

lazypower:
+1 LGTM, thanks for this! :beers:

PR #23, opened 2016-06-08
Making the ports mutable by using member update and rendering default file

mbruzek:


reactive/etcd.py
```diff
@@ -341,6 +392,16 @@ def passive_dismiss_context(cluster):
     cluster.dismiss()
 
 
+def close_open_ports():
+    ''' Close the previous port and open the port from configuration. '''
+    configuration = hookenv.config()
+    previous_port = configuration.previous('port')
+    port = configuration.get('port')
+    if previous_port != port:
+        close_port(previous_port)
```
lazypower, line 414: You had made mention this was of concern and its a legitimate concern. On first run it is indeed None.

subprocess.CalledProcessError: Command '['close-port', 'None/TCP']' returned non-zero exit status 2

This can be avoided with a quick fix to the inline if statement:

``` python
if previous_port != port and previous_port is not None:
```

templates/systemd
```diff
@@ -1,17 +1,19 @@
...
 [Service]
 Environment=DAEMON_ARGS=
 Environment=ETCD_NAME=%H
-Environment=ETCD_DATA_DIR=/var/lib/etcd/
+Environment=ETCD_DATA_DIR=/var/lib/etcd/default
 EnvironmentFile=-/etc/default/%p
 Type=notify
 User=etcd
 PermissionsStartOnly=true
+#ExecStart=/bin/sh -c "GOMAXPROCS=$(nproc) /usr/bin/etcd $DAEMON_ARGS"
```
lazypower, line 16: nit: if we're going to leave this in, we should probably make it tuneable with a default to nproc and uncomment it.

But that can be done in another PR.

reactive/etcd.py
```diff
@@ -57,6 +59,54 @@ def remove_states():
     remove_state('etcd.installed')
 
 
+@when('etcd.installed')
+@when('leadership.is_leader')
+@when_any('config.changed.port', 'config.changed.management_port')
+def leader_config_changed():
+    ''' '''
```
lazypower: Kittens have cried over this attempt at a docstring 😿

lib/etcdctl.py
```diff
@@ -19,11 +19,12 @@ def register(self, cluster_data):
...
         management_port
         '''
-        command = "etcdctl -C {} member add {}" \
-                  " http://{}:{}".format(cluster_data['leader_address'],
-                                         cluster_data['unit_name'],
-                                         cluster_data['private_address'],
-                                         cluster_data['management_port'])
+        # Create a https url to the leader unit name on the private addres.
+        command = "etcdctl -C {0} member add {1} " \
+                  "https://{2}:{3}".format(cluster_data['leader_address'],
```
lazypower, line 24: CR :cop: approves of this bug squash

Fixes #24

lazypower:
Once the inline patch left in the comments has been merged,  and you fill out the docstring for `leader_config_changed` this PR is ready to ⛵ 

i've deployed, scaled, and changed config on LXD  as verification

lazypower:
This PR +1 LGTM

PR #25, opened 2016-06-09
Minor fixes

lazypower:
Leader_sets teh address when port has changed
re-arranges the parameters when using multi-conversation objects, the stacking of @whens is applied in reverse order


mbruzek:
+1 lgtm

PR #26, opened 2016-06-09
Minor update to sending ssl credentials

lazypower:
The interface method was renamed to set_client_credentials and now additionally hands off the CA for services that require it (like swarm)


mbruzek:
This change looks fine and related to changes in the interface

PR #27, opened 2016-06-15
Refactor tests to be TLS friendly

lazypower:
Funny thing, TLS works for cluster health "implicitly" but not during member list. This may warrant further investigation but the tests cleanly pass on AWS, GCE, MAAS, and LXD providers.

### DIFF
--- a/tests/10-deploy
+++ b/tests/10-deploy
@@ -7,30 +7,58 @@ import unittest
 class TestDeployment(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        cls.deployment = amulet.Deployment(series='xenial')
-        cls.deployment.add('etcd')
-        try:
-            cls.deployment.setup(timeout=900)
-            cls.deployment.sentry.wait()
-        except amulet.helpers.TimeoutError:
-            msg = "Environment wasn't stood up in time"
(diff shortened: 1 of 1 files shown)


arosales:
All reports I have heard have been positive. Thanks for the work on this @chuckbutler

tests/10-deploy
```diff
@@ -7,30 +7,61 @@ import unittest
...
+        for unit in self.etcd:
+            status = unit.run('service etcd status')
+            self.assertFalse(status[1] == 1)
+            self.assertTrue("running" in status[0])
+
+    def test_cluster_health(self):
+        ''' Iterate all the units and verify we have a clean bill of health
+        from etcd '''
+        for unit in self.etcd:
+            health = unit.run('etcdctl cluster-health')
```
mbruzek, line 43: we should call health.lower() here. Otherwise "Unhealthy" will not mach "unhealthy" right?

mbruzek, line 43: disregard, it looks like all keywords are lowercase in the output.

lazypower, line 43: That doesn't mean they won't change the message and give us a red herring. Good call on this! 👍

PR #32, opened 2016-06-30
Adds missing implementation of etcd-proxy

lazypower:
The proxy relation now properly sends cluster peer details, and client connection credentials.

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -152,12 +152,26 @@ def send_single_connection_details(db):
     # Set the connection string on the db relation.
     db.set_connection_string(connection_string)
 
-# #
-# @when('proxy.connected')
-# def send_cluster_details(proxy):
-#     etcd = EtcdHelper()
-#     proxy.provide_cluster_string(etcd.cluster_string())
-#
+
+@when('proxy.connected')
+def send_cluster_details(proxy):
+    cert = leader_get('client_certificate')
+    key = leader_get('client_key')
+    ca = leader_get('certificate_authority')
+    proxy.set_client_credentials(key, cert, ca)
+
+    # format a list of cluster participants
+    etcdctl = EtcdCtl()
+    peers = etcdctl.member_list()
+    cluster = []
+    for peer in peers:
+        # Potential member doing registration. Default to skip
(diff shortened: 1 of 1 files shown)


nelljerram:
Thanks - LGTM.

reactive/etcd.py
```diff
@@ -152,12 +152,27 @@ def send_single_connection_details(db):
...
+    key = leader_get('client_key')
+    ca = leader_get('certificate_authority')
+    proxy.set_client_credentials(key, cert, ca)
+
+    # format a list of cluster participants
+    etcdctl = EtcdCtl()
+    peers = etcdctl.member_list()
+    cluster = []
+    for peer in peers:
+        thispeer = peers[peer]
```
mbruzek, line 168: Isn't "peer" the same as "thispeer" in this case? Or am I missing a subtlety?

lazypower, line 168: peer becomes the string key of the dict.

'etcd0', so i use it as a scope operator for the dict. We can revise that for better clarity, as i had the same incorrect assumption initially

lazypower:
@mbruzek was there something we wanted to do with this implementation? @neiljerram has given me a sign off in IRC that it works for their purposes.

reactive/etcd.py
mbruzek, line 168: for peer in peers: will create an iterator that makes the peer variable every member of the peers list. 
thispeer = peers[peer] is the same thing as peer is it not?

mbruzek:
@chuckbutler I have no problems with this change, the for loop looks a little strange to me, I still think peer and thispeer are equivalent and I don't understand the difference. That looks like unnecessary code to me, but if there is a subtlety that I am missing, a comment would clear that up.

reactive/etcd.py
lazypower, line 168: I encourage you to run that, debug it, and tell me what you see. I'm consistently seeing it evaluate to the key, and not the dict of values contained at that key.

mbruzek:
OK I am finally +1 to this change. 

`etcdctl.member_list()` actually returns a dictionary (not a list as the method name would imply), so the iterator is on the keys and you need to get the peer member by the key name.

PR #36, opened 2016-08-02
Remove trusty support from the layer

lazypower:
Trusty doesn't have an archive option for installing etcd. The big
line items this resolves is crossing the barrier from tls by default.

Trusty series published versions of this charm have all been targeting
development channels. I feel this is a good time to break that backwords
compat and move forward with tls by default in xenial.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -12,7 +12,6 @@ tags:
   - layer
 series:
   - xenial
-  - trusty
 min-juju-version: 2.0-beta6
 resources:
   etcd:


mbruzek:
What specifically changed to make the latest etcd charm incompatible with the trusty charm? We should also update the readme with this information (directory where the files are stored, or whatever), and how users should turn up a new etcd cluster if they were using the old incompatible one.

lazypower:
Good insight, i'll get that added an commit an update.

lazypower:
superseded by #37

PR #39, opened 2016-08-08
Tweaks for automated testing

lazypower:
Removes the 00-setup in favor of tests.yaml
Adds timeout values to add_unit, and bumps them up for slow clouds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: python
+python:
+ - '3.5'
+
+install:
+ - pip install tox
+
+script:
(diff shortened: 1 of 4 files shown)


mbruzek:
It looks to me like you increased the amount of time to wait for a successful deployment. :watch: :cop: does not agree with changes based on people mis-understanding of time.

People assume that time is a strict progression of cause to effect, but actually from a non-linear, non-subjective viewpoint - it's more like a big ball of wibbly wobbly... time-y wimey... stuff.

PR #44, opened 2016-09-15
Adds the application version set stanza

lazypower:
Adds the application version to the juju status output

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -16,6 +16,7 @@ from charmhelpers.core.hookenv import is_leader
 from charmhelpers.core.hookenv import leader_set
 from charmhelpers.core.hookenv import leader_get
 
+from charmhelpers.core.hookenv import application_version_set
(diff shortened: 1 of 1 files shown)


reactive/etcd.py
```diff
@@ -51,6 +53,14 @@ def check_cluster_health():
     status_set('active', health['status'])
 
 
+@when('etcd.installed')
+def set_app_version():
+    # sneak in the version as well
+    cmd = ['etcd', '-version']
+    version = check_output(cmd).split(b'\n')[0].split(b':')[-1]
```
mbruzek, line 60: This change looks fine to me, but I would recommend putting the version format that is **currently** returned in the comment above the split to show the developer what you expected and why it works. That way if the version ever changed in a way that breaks this logic we can begin to see what you had expected and what it returns now.

mbruzek:
+1 from me on this one.

PR #47, opened 2016-09-19
Cleanup messaging during install

lazypower:
remove attempting line, as it doesn't inspire confidence


mbruzek:
+1 to combining 2 status messages to one.

PR #48, opened 2016-09-22
strip the extra space from version output

lazypower:
Adds better comment around the method so its more obvious when this breaks


mbruzek:
+1 thanks for adding the version output so we can figure out the changes that are necessary.

PR #49, opened 2016-09-22
Replace ltrim with lstrip

lazypower:
Python != PHP


mbruzek:
+1 for the more _pythonic_ lstrip()

PR #52, opened 2016-09-23
Improvements on the status messaging

lazypower:
This replaces etcd's normal health message with an improved health
status and peer unit count in the message


reactive/etcd.py
```diff
@@ -50,7 +50,21 @@ def check_cluster_health():
...
+    else:
+        unit_health = "Unhealthy"
+
+    # Determine units peer count, and surface 0 by default
+    try:
+        peers = len(etcdctl.member_list())
+    except Exception:
+        peers = 0
+
+    status_message = "{0} with {1} known peers".format(unit_health, peers)
```
mbruzek, line 66: s/known peers/total members/ ?

PR #53, opened 2016-09-25
Make messaging uniform again

lazypower:
Correct messaging with full stop short sentences.


mbruzek:
These status messages look uniform! +1

PR #55, opened 2016-10-13
Moved 10-deploy to 10-deploy.py

lazypower:
The default behavior of the sentry.wait() method was causing some false
positives in CWR. I've switched over to wait_for_messages() to sniff out
the ETCD deployment status and only proceed once Healthy/Unhealthy has
been reported.

One might argue why I didn't only decorate for healthy. I feel that the
output of the tests is still going to be helpful in determining the
severity of the failure, and might give insight into why the deployment
is unhealthy.

### DIFF
--- a/tests/10-deploy.py
+++ b/tests/10-deploy.py
@@ -2,6 +2,7 @@
 
 import amulet
 import unittest
+import re
 
 
 class TestDeployment(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


tests/10-deploy.py
```diff
@@ -10,7 +11,9 @@ def setUpClass(cls):
         cls.d = amulet.Deployment(series='xenial')
         cls.d.add('etcd')
         cls.d.setup(timeout=1200)
-        cls.d.sentry.wait()
+        cls.d.sentry.wait_for_messages({'etcd':
+                                        re.compile('Healthy*|Unhealthy*')})
```
mbruzek, line 15: I have not used this method before. It will take a regular expression? Sure enough:

> A mapping of services to an exact message, a regular expression, a set of messages or regular expressions, or a list of messages or regular expressions. 

Reference: https://pythonhosted.org/amulet/amulet.html#amulet.sentry.Talisman.wait_for_messages

Thanks @johnsca as this was no doubt your excellent work with regular expressions!

mbruzek:
Great work here @chuckbutler on improving the tests! I learned something that will help me write better tests. +1

PR #56, opened 2016-10-13
Dont reset env when testing

lazypower:


mbruzek:
+1 to this clean up. Thanks @chuckbutler

PR #57, opened 2016-10-13
Removing an ignore override

lazypower:


mbruzek:
I forgot we copied that file here, +1 for removing it and I landed a fix in the tls layer where it will be excluded.

PR #60, opened 2016-11-08
Strip out the (leader) status message appender

lazypower:
Leader status is now represented as an asterisk next to the unit name,
negating the need to carry this identifier forward.

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -15,6 +15,7 @@ from charmhelpers.core.hookenv import log
 from charmhelpers.core.hookenv import is_leader
 from charmhelpers.core.hookenv import leader_set
 from charmhelpers.core.hookenv import leader_get
+from charmhelpers.core.hookenv import status_set
 
 from charmhelpers.core.hookenv import application_version_set
 from charmhelpers.core.hookenv import open_port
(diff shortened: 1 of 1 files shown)


Cynerva:
Mind taking out the old import at [line 13](https://github.com/chuckbutler/layer-etcd/blob/65541447a70458227cbe00794211c010dc6e822d/reactive/etcd.py#L13)?

`from charmhelpers.core.hookenv import status_set as hess`

lazypower:
Ah, i thought i had gotten that scrubbed, good eye

reactive/etcd.py
```diff
@@ -446,11 +447,3 @@ def close_open_ports():
...
     ''' This method wraps the bash "install" command '''
     return check_call(split('install {} {}'.format(src, tgt)))
-
-
-def status_set(status, message):
-    ''' This is a fun little hack to give me the leader in status output
-        without taking it over '''
-    if is_leader():
-        message = "{} (leader)".format(message)
-    hess(status, message)
```
wwwtyro, line 456: Can we get rid of the `import blah blah as hess` now?

wwwtyro: oops, shoulda read cynerva first

lazypower:
Ok, scrubbed that funky import that needed to go die in a :fire: 

Thanks for catching that. I thought i got all traces but i see the import was above the fold in my editor... doh.

Cynerva:
Cool. Thanks for cleaning this up, by the way!

PR #63, opened 2016-11-16
Adds initial persistent external storage support

lazypower:
To use on AWS EBS disk type:

juju deploy etcd
juju add-storage etcd/0 data=ebs,10g

To use on GCE PD Disk type:

juju deploy etcd
juju add-storage etcd/0 data=gce,10g

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -455,13 +455,10 @@ def format_and_mount_storage():
     if os.path.isdir('/var/lib/etcd/default'):
         hookenv.log('Detected existing data, migrating to new location.')
         # Migrate any existing data
-        cmd = ['mkdir', '-p', '/mnt/etcd-migrate']
-        hookenv.log('With command: {}'.format(' '.join(cmd)))
-        check_call(cmd)
-
+        os.makedirs('/mnt/etcd-migrate', exist_ok=True)
         mount_volume(block, '/mnt/etcd-migrate')
 
-        cmd = ['rsync', '-avz', '/var/lib/etcd/default', '/mnt/etcd-migrate/']
+        cmd = ['rsync', '-azp', '/var/lib/etcd/default', '/mnt/etcd-migrate/']
         hookenv.log('With command: {}'.format(' '.join(cmd)))
         check_call(cmd)
 
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -15,6 +15,7 @@ from charmhelpers.core.hookenv import is_leader
 from charmhelpers.core.hookenv import leader_set
 from charmhelpers.core.hookenv import leader_get
 from charmhelpers.core.hookenv import status_set
+from charmhelpers.core.hookenv import storage_get
 
 from charmhelpers.core.hookenv import application_version_set
 from charmhelpers.core.hookenv import open_port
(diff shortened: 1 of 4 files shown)


marcoceppi:
README update would be nice. a few common callouts, like you have for AWS (ebs, ebs-ssd), GCE, then link to the Juju documentation.

Have not done a code review yet

reactive/etcd.py
```diff
@@ -431,6 +432,87 @@ def passive_dismiss_context(cluster):
...
+    cmd = ['mkfs.ext4', device_info['location'], '-F']
+    hookenv.log('Creating filesystem on {}'.format(device_info['location']))
+    hookenv.log('With command: {}'.format(' '.join(cmd)))
+    check_call(cmd)
+
+    # Only attempt migration if directory exists
+    if os.path.isdir('/var/lib/etcd/default'):
+        hookenv.log('Detected existing data, migrating to new location.')
+        # Migrate any existing data
+        cmd = ['mkdir', '-p', '/mnt/etcd-migrate']
```
marcoceppi: Lets use the python module for this: https://docs.python.org/3/library/os.html?highlight=makedirs#os.makedirs

reactive/etcd.py
```diff
@@ -431,6 +432,87 @@ def passive_dismiss_context(cluster):
...
+        hookenv.log('With command: {}'.format(' '.join(cmd)))
+        check_call(cmd)
+
+        mount_volume(block, '/mnt/etcd-migrate')
+
+        cmd = ['rsync', '-avz', '/var/lib/etcd/default', '/mnt/etcd-migrate/']
```
marcoceppi: `v` is excessive and `p` (partials) would be a good flag to add. If rsync is interrupted it'll pick up partials faster

reactive/etcd.py
```diff
@@ -431,6 +432,87 @@ def passive_dismiss_context(cluster):
...
+        hookenv.log('With command: {}'.format(' '.join(cmd)))
+        check_call(cmd)
+
+        unmount_path('/mnt/etcd-migrate')
+
+    # halt etcd to perform the data-store migration
+    host.service_stop('etcd')
```
marcoceppi, line 468: Do we need to perform a second, quick sync after service_stop in the event data has changed since rsync and before halting?

marcoceppi (changes requested):

mbruzek:
I upgraded the etcd charm with a charm that I built from this branch. It upgraded cleanly and I was able to attach storage, as indicated in the top step. I do think we need the storage steps to the README.md file.

```
$ juju storage etcd
[Storage]  
Unit       Id      Location   Status    Message  
etcd/0     data/0  /dev/xvdf  attached 
```

```
# df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            1.9G     0  1.9G   0% /dev
tmpfs           377M   11M  367M   3% /run
/dev/xvda1      7.8G  2.0G  5.4G  28% /
tmpfs           1.9G     0  1.9G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           1.9G     0  1.9G   0% /sys/fs/cgroup
/dev/xvdf       9.8G  341M  8.9G   4% /var/lib/etcd
tmpfs           377M     0  377M   0% /run/user/1000
```

LGTM +1

reactive/etcd.py
lazypower: well, i could just move the service stop up. I was thinking the less time the service is offline the better, but i can see your point here, that we might be missing a few stray key-val sets.

PR #64, opened 2016-11-17
Fixing the source with the new flake8 rules.

mbruzek:
Using flake8 version 3.2.0


reactive/etcd.py
```diff
@@ -443,7 +442,7 @@ def format_and_mount_storage():
 
     if volume_is_mounted(block):
         hookenv.log('Device is already attached to the system.')
-        hookenv.log('Cowardly refusing to take action against {}'.format(block))
+        hookenv.log('Refusing to take action against {}'.format(block))
```
lazypower, line 445: I dont think this was necessary

Issue #66, opened 2016-12-07
Fix travis failures in etcd


mbruzek:
The travis.yml calls tox and the tox tests are broken. We should fix this.

chuckbutler:
I'm +1 to this notion. 

The charm will have to be built and the tests will need to be executed against the assembled charm unless there's a way to stub out LAYER_OPTIONS. Thats the root of the failure from the change set.

mbruzek:
@chuckbutler I see that you reopened this. I see some commits after this merge, but I don't see a PR. Do we need to do something else here?

chuckbutler:
I'm not really happy with the implementation. it works, but its a gigantic bandaid covering up a larger issue which is how do you test isolated layers. We can close this and open a new issue to track fixing the hack.

PR #67, opened 2016-12-07
Fixes for #66

lazypower:
Moves the install routine to the travis yaml, and builds to a temporary
path in PWD, then executes the tox tests


mbruzek:
brilliant solution to this weird problem +1

PR #73, opened 2017-01-25
add and nrpe-external-master relation, with a basic check (systemd service is running)

axinojolais:
Yup.


axinojolais:
Will resubmit

PR #74, opened 2017-01-25
add an nrpe-external-master relation, with a basic check (systemd service is running)

axinojolais:
Relates to https://github.com/juju-solutions/bundle-canonical-kubernetes/issues/165

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -25,6 +25,7 @@ from charmhelpers.core import hookenv
 from charmhelpers.core import host
 from charmhelpers.fetch import apt_update
 from charmhelpers.fetch import apt_install
+from charmhelpers.contrib.charmsupport import nrpe
 
 from etcdctl import EtcdCtl
 from etcdctl import get_connection_string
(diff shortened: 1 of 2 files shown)


reactive/etcd.py
```diff
@@ -505,6 +506,30 @@ def read_tls_cert(cert):
...
+def update_nrpe_config(unused=None):
+    # List of systemd services that will be checked
+    services = ('etcd',)
+
+    # The current nrpe-external-master interface doesn't handle a lot of logic,
+    # use the charm-helpers code for now.
+    hostname = nrpe.get_nagios_hostname()
+    current_unit = nrpe.get_nagios_unit_name()
+    nrpe_setup = nrpe.NRPE(hostname=hostname, primary=False)
+    nrpe.add_init_service_checks(nrpe_setup, services, current_unit)
```
mbruzek, line 529: It looks like this method has a very permissive reactive clause. Is this add idempotent? Will multiple adds work?

reactive/etcd.py
```diff
@@ -505,6 +506,30 @@ def read_tls_cert(cert):
...
 
+@when('nrpe-external-master.available')
+@when_not('nrpe-external-master.initial-config')
+def initial_nrpe_config(nagios=None):
+    set_state('nrpe-external-master.initial-config')
+    update_nrpe_config(nagios)
+
+
+@when('etcd.installed')
+@when('nrpe-external-master.available')
```
mbruzek, line 517: You also need to include logic to remove the nrpe check when the relationship to etcd is removed/destroyed.

reactive/etcd.py
```diff
@@ -25,6 +25,7 @@
 from charmhelpers.core import host
 from charmhelpers.fetch import apt_update
 from charmhelpers.fetch import apt_install
+from charmhelpers.contrib.charmsupport import nrpe
```
mbruzek, line 28: Is this library provided by the nrpe interface?

mbruzek (changes requested):
Thanks for the contribution, I honestly appreciate it. 

We do have to worry about add and remove relations so having checks for removing the relation are needed here.

reactive/etcd.py
axinojolais: No, it's in charmhelpers, which is provided by hmmm the basic layer I think ? I'm not sure I understand your question.

reactive/etcd.py
axinojolais: Indeed, working on that.

reactive/etcd.py
axinojolais: A "very permissive reactive clause" ? AIUI, it will only fire when the config option "nagios_context" or "nagios_servicegroups" is changed, which shouldn't happen all that often. Is that not the case ?

axinojolais:
OK, the check will get removed upon nrpe-external-master departure with the last commit.

lazypower:
+1 LGTM

Thanks for the submission @axinojolais. If you don't get to the re-sync i'll try to pull it down and rebase for you. Thanks again!

lazypower (approved):

axinojolais:
@mbruzek @chuckbutler I just resolved the (trivial) conflict, this should be good to merge :)

Thanks

reactive/etcd.py
marcoceppi: 👍 this is included in charmhelpers

marcoceppi (approved):

lazypower:
Thanks for the contribution! This will ship with the next release of the etcd charm

PR #76, opened 2017-01-30
Add debug action

Cynerva:
Follow-up to a recent debug session. Let's get the debug action added to layer-etcd.

### DIFF
--- a/debug-scripts/etcd
+++ b/debug-scripts/etcd
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -ux
 
-systemctl status etcd > etcd-systemctl-status
+systemctl status etcd > $DEBUG_SCRIPT_DIR/etcd-systemctl-status
 journalctl -u etcd > $DEBUG_SCRIPT_DIR/etcd-journal
 
 alias etcdctl="etcdctl --cert-file /etc/ssl/etcd/client.crt --key-file /etc/ssl/etcd/client.key --ca-file /etc/ssl/etcd/ca.crt"
--- a/debug-scripts/etcd
+++ b/debug-scripts/etcd
@@ -3,6 +3,8 @@ set -ux
 
 systmctl status etcd > etcd-systemctl-status
 journalctl -u etcd > $DEBUG_SCRIPT_DIR/etcd-journal
+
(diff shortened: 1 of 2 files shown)


Cynerva:
Fixed a couple problems in testing, oops. This should be good to go now.

debug-scripts/etcd
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -ux
+
+systemctl status etcd > $DEBUG_SCRIPT_DIR/etcd-systemctl-status
+journalctl -u etcd > $DEBUG_SCRIPT_DIR/etcd-journal
+
+alias etcdctl="etcdctl --cert-file /etc/ssl/etcd/client.crt --key-file /etc/ssl/etcd/client.key --ca-file /etc/ssl/etcd/ca.crt"
+etcdctl cluster-health > $DEBUG_SCRIPT_DIR/etcdctl-cluster-health
+etcdctl ls --recursive > $DEBUG_SCRIPT_DIR/etcdctl-ls
+etcdctl member list > $DEBUG_SCRIPT_DIR/etcdctl-member-list
```
Cynerva, line 10: @chuckbutler I made a naive attempt to grab some useful info from etcd. Do you have any ideas what else we might want here?

lazypower: can we dump x509 info for the tls certificate(s)? I'm still thinking through what other diagnostic information would be helpful, but this is a great start @Cynerva <3

[edit] acutally, does layer:tls-client have the tls debug bits added? it might be better to place it there so every  configured tls-client has this out of the box.

Cynerva: Sure thing.

Hey, actually, any objection to me submitting that to [layer-tls-client](https://github.com/juju-solutions/layer-tls-client) instead? That'll get it included in the kubernetes charms, too.

I'm guessing that'd be useful? I dunno!

lazypower: Certainly. I think it makes sense to live there so long as we are parsing out the locations from the layer.yaml.

It became a little more obvious to me the more I thought about this that we're going to have to rely on the implementor to actually supply the TLS locations for x509 validation. Thats potentially problematic as I don't know that we can reasonably infer what layer it was included in.

I think it needs more thought and perhaps we can get a dict back of the layer options and just iterate over them to discover which ones have tls-cert keys and can then run the gambit of validations there.

Thanks @Cynerva for the submission. I'm good with this as is and will gladly take further patches to enhance functionality.

PR #77, opened 2017-02-19
WIP - Working multi-snap cluster

lazypower:
DO NOT MERGE!

This is an open thread to track the progress and discuss the approach to the refactoring of the etcd layer to migrate from apt delivery => snaps.

This has implications on:

- delivery mechanism
- backup/restore
- durable storage migration(s)

Initial deployment + upgrade code. Functional X node cluster. Does not survive leader termination.


templates/defaults
```diff
@@ -1,5 +1,5 @@
 # This file is rendered by Juju, manual edits will not be persisted
-ETCD_DATA_DIR=/var/lib/etcd/default
+ETCD_DATA_DIR={{ etcd_data_dir }}/{{ unit_name }}.etcd
```
lazypower, line 2: This now matches what the snap is doing. I'm not 100% positive we want to keep this nomenclature but I'm for keeping it in lockstep with the snap so it stays consistent.

reactive/etcd.py
```diff
@@ -61,21 +60,25 @@ def check_cluster_health():
     except Exception:
         peers = 0
 
-    status_message = "{0} with {1} known peers.".format(unit_health, peers)
+    if peers > 1:
+        subject = "peers"
+    else:
+        subject = "peer"
+
+    status_message = "{0} with {1} known {2}.".format(unit_health, peers,
```
marcoceppi: Could we do this instead?

```python
status_message = '{0} with {1} known peer{2}'.format(unit_health, peers, 's' if peers != 1 else '')
```

lazypower: We could, I don't know what it buys us other than collapsing four lines into one, but at the sake of readability. I don't feel strongly either way however.

actions/snap-upgrade
```diff
@@ -0,0 +1,53 @@
...
+# Handle the snap ingestion routine
+function ingest_snap {
+  # This determines if we have a resource provided etcd ingest snap.
+  if [ "${USE_RESOURCE}" == "False" ]; then
+    # If no manual snap specified, assume snapstore connectivity
+    snap install etcd --channel=ingest/stable --classic
+  else
+    # Install from the resource location, and check for correct snap.
+    MANUAL_SNAP=$(resource-get etcd)
+    snap install ${ETCD_MANUAL_SNAP} --classic
```
adam-stokes: If you are installing a local snap you need to pass `--dangerous` as well

actions/snap-upgrade
```diff
@@ -0,0 +1,53 @@
...
+if [ ! $(charms.reactive is_state etcd.snap.ingest) ]; then
+  ingest_snap
+fi
+
+if [ "${USE_RESOURCE}" == "False" ]; then
+  CONFIGURED_CHANNEL=$(config-get channel)
+  snap refresh etcd --channel=${CONFIGURED_CHANNEL} --classic
+else
+  MANUAL_SNAP=$(resource-get etcd)
+  snap refresh ${MANUAL_SNAP}
```
adam-stokes: Also i dont think snap refresh works for locally installed snaps

marcoceppi: You'll need to just snap install /again/

lazypower:
Closing for a formal PR

PR #78, opened 2017-03-09
Snap Support (etcd 2.x/3.x forward facing)

lazypower:
This introduces a large refactoring of the layer codebase to be compliant with Snaps and the data-paths for Snap packages. The charm strives to use strict confinement primitives and enable the snaps to do complex operations at scale using only template logic and minimal intervention.

Notable Updates in this branch:
- Introduces a manual path for snap-upgrade, to ensure the operator is aware of this change
- Migration path assumes the snap-store, or manual attachment of snaps in offline environment(s)
- Refactors external data mounts to reside in /media per the 'removable-media' interface in snaps.
- Supports both etcd 3.x and 2.x
- refactors critical-path actions to python from bash

Note: in order for this branch to be properly used, the etcd-snap will require  https://github.com/markshuttle/etcd-snaps/pull/4  before any external storage can be used. Snap confinement will block the process otherwise.


actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+
+def unpack_resource():
+    ''' Grab the resource path, and unpack it into $PATH '''
+    cmd = "tar xvf {0} -C {1}".format(SNAPSHOT_ARCHIVE, ETCD_DATA_DIR)
+    check_call(split(cmd))
+
+
+def start_etcd_forked():
+    ''' Start the etcd daemon temporarily to initiate new cluster details '''
+    raw = "/snap/etcd/current/bin/etcd -data-dir={0} -force-new-cluster"
```
wwwtyro, line 72: Does etcd recursively create the data-dir if it doesn't exist?

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+    cmd = raw.format(ETCD_DATA_DIR)
+    Popen(split(cmd), stdout=PIPE, stderr=PIPE)
+
+
+def pkill_etcd():
+    ''' Kill the temporary forked etcd daemon '''
+    cmd = 'pkill etcd'
```
wwwtyro: Can we kill by PID instead? Little nitpicky, but feels better. I think the Popen object includes the pid.

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+    check_call(split(cmd))
+
+
+def probe_forked_etcd():
+    ''' Block until the forked etcd instance has started and return'''
+    output = b""
+    loop = 0
+    MAX_WAIT = 10
```
wwwtyro, line 93: Do you think this should be configurable?

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+            data = fp.read(BUF_SIZE)
+            if not data:
+                break
+            shasum.update(data)
+    return shasum.hexdigest()
+
+
+if __name__ == '__main__':
+    preflight_check()
+    stop_etcd()
```
wwwtyro, line 150: It doesn't look like we're bailing here if we fail to stop etcd. Do we need to?

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+        log('Failed to stop snap.etcd.etcd')
+
+    try:
+        service_stop('etcd')
+        log('Stoppd service: etcd')
+    except:
+        log('Failed to stop service: etcd')
+
+
+def shasum_file(filepath):
```
wwwtyro, line 135: Really nicely done. No memory issues, eh? :+1:

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+
+def render_backup():
+    ''' Backup existing data in the event of restoration on a dirty unit. '''
+    if not os.path.isdir(ETCD_DATA_DIR):
+        return
+
+    with chdir(ETCD_DATA_DIR):
+        if not SKIP_BACKUP:
+            log('Backing up existing data found in {}'.format(ETCD_DATA_DIR))
+            # tar cvf $TARGET_PATH/$ARCHIVE default
```
wwwtyro: Does this need to happen?

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+            loop = loop + 1
+        time.sleep(1)
+        if loop > MAX_WAIT:
+            raise TimeoutError("Timed out waiting for forked etcd.")
+
+
+def reconfigure_client_advertise():
+    ''' Reconfigure the backup to use host network addresses for client advertise
+        instead of the assumed localhost addressing '''
+    cmd = "/snap/bin/etcd.etcdctl member list"
```
wwwtyro, line 111: Nitpicky, but you use this string twice (other instance on line 87). No objection if you want to define a variable for it.

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+def preflight_check():
+    ''' Check preconditions for data restoration '''
+    if not SNAPSHOT_ARCHIVE:
+        action_fail({'result.failed': 'Missing snapshot. See: README.md'})
+        sys.exit(0)
+
+
+def render_backup():
+    ''' Backup existing data in the event of restoration on a dirty unit. '''
+    if not os.path.isdir(ETCD_DATA_DIR):
```
wwwtyro: Since we're setting SKIP_BACKUP to default true, should we log if there's no data dir and SKIP_BACKUP is false?

actions/snap-upgrade.py
```diff
@@ -0,0 +1,144 @@
...
+
+    if os.path.exists(backup_package):
+        msg = 'Backup package exists: {}'.format(backup_package)
+        action_set({'fail.message': msg})
+        return False
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        # Create a temporary path to perform the backup, and date the contents.
+        dated_path = "{0}/etcd_migration_{1}".format(tmpdir, datestring)
+        os.makedirs(dated_path, exist_ok=True)
```
wwwtyro: Is it really okay if this already exists? Seems like it would be a weird case?

actions/snap-upgrade.py
```diff
@@ -0,0 +1,144 @@
...
+            action_fail('Migration failed')
+
+
+def purge_deb_files():
+    log('Purging deb configuration files post migration', 'INFO')
+    cmd = 'apt purge etcd'
+    check_call(split(cmd))
+    for f in deb_paths['config']:
+        log('Removing file {}'.format(f), 'INFO')
+        os.remove(f)
```
wwwtyro: I think this throws an error if the file doesn't exist. Do you want to deal with that here?

wwwtyro:
@chuckbutler I left few minor comments, but this looks good, per usual. Nice work. :+1:

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+
+ETCD_DATA_DIR = opts['etcd_data_dir']
+ETCD_PORT = config('management_port')
+PRIVATE_ADDRESS = unit_get('private-address')
+SKIP_BACKUP = action_get('skip-backup')
+SNAPSHOT_ARCHIVE = resource_get('snapshot')
+TARGET_PATH = action_get('target')
+
+
+def preflight_check():
```
mbruzek, line 38: Since we are going to snaps do you need to add checks if you can write in the "etcd_data_dir" or something of that nature? I see that is a layer option and if someone did not know about snaps or confinement they may try to configure the directory incorrectly.

actions/restore.py
mbruzek: Good use of the hashlib here. @wwwtyro 's comment makes me think about shelling out to sha256sum in this case due to memory concerns. But you do whichever you want.

actions/restore.py
```diff
@@ -0,0 +1,147 @@
...
+        while True:
+            data = fp.read(BUF_SIZE)
+            if not data:
+                break
+            shasum.update(data)
+    return shasum.hexdigest()
+
+
+if __name__ == '__main__':
+    preflight_check()
```
mbruzek, line 149: This is awesome code! Very clear steps that are easy to follow. Wondering if it is worth it printing each step out so the user knows what is going or how each step went.

reactive/etcd.py
```diff
@@ -155,30 +140,28 @@ def leader_config_changed():
             # Update the member's peer_urls with the new ports.
             log(etcdctl.member_update(members[unit_name]['unit_id'], url))
         # Render just the leaders configuration with the new values.
-        render('defaults', '/etc/default/etcd', bag.__dict__, owner='root',
-               group='root')
+        render_config()
```
mbruzek, line 143: You are rendering configuration here and _not_ passing the bag variable. Did you mean to do that or not? I don't see bag getting any values assigned in this method but the other uses appear to take the bag as a parameter.

mbruzek (approved):
I looked over this code it looks very clean and clear! Really liked how the code reads, well done sir! None of my comments or questions are blockers. 

+1 from me. Looking forward to using the new etcd.

actions/restore.py
wwwtyro: Ah, my bad. I was trying to communicate that this intelligently takes memory into account, and so shouldn't have a problem. Clearly, I failed!

actions/restore.py
marcoceppi: I suppose so, but since layer options are a compile time flag and not a runtime normal users aren't exposed to it.

actions/restore.py
mbruzek: If the etcd data > total RAM I would worry about memory. I just don't know how large the etcd data is.

wwwtyro: @mbruzek Yeah exactly! That's why I like this, he's chunking it into 64KB blocks. :)

mbruzek: My understanding of the `update()` method is wrong, I withdraw my concern and comment. Great job here Chuck.

actions/restore.py
lazypower: Oh man I'm so happy we have code review. Whoops!

marcoceppi (changes requested):
Great start, I think some of the feedback needs to be addressed

reactive/etcd.py
lazypower: Passing in Bag is only really necessary when you have changes to the bag object (like updating a cluster string with a seeded value from leader_data, as seen in line: 315. 

Good call out, but I think this is an acceptable time to use the objects probed defaults.

lazypower:
@marcoceppi  This is ready for a re-review unless you feel any of the skipped comments are critical path.

actions/restore.py
lazypower: Not necessarily. As its working with snap confinement, it can and will be blocked if its configured outside of /media or $SNAP_DATA/$SNAP_COMMON/$SNAP_USER

mbruzek:
I have deployed the old etcd charm and ran through the README.md instructions. I found some breakages that need to to be addressed and spoke with @chuckbutler in a hangout watching him fix the problems I encountered.  I will give it another review tomorrow when the work we discussed lands.

ktsakalozos:
Probably not related to this PR but I got into trouble with nvidia drivers: http://pastebin.ubuntu.com/24215186/

It could be I am missing some update: https://bugs.launchpad.net/snap-confine/+bug/1665592

lazypower:
@ktsakalozos was this on lxd?

There is a known breakage in lxd when using snap refresh. xref here https://bugs.launchpad.net/snapd/+bug/1668659

ktsakalozos:
@chuckbutler yeap it was on lxd

lazypower:
While not excellent, this makes me feel better that we are aware of this issue and have I tracked in two separate issues targeting specifics of the behavior illustrated.   I'm hacking on etcd upgrade tests in amulet/bundletester wizardry now so we know we don't break this moving forward.

ktsakalozos:
One-liner fix for our bundle tests https://github.com/juju-solutions/bundle-canonical-kubernetes/pull/239 so they look for etcd binary under snap/bin

mbruzek:
The latest revision also passes my review. All checks pass. Let me merge this. If there are any other improvements we should open separate bugs.

PR #80, opened 2017-03-22
Adjust pathing for tls/bins for snap deployment

lazypower:


mbruzek:
Thanks for updating this test @chuckbutler bringing the xenial systemd goodness.

+1 from me

PR #81, opened 2017-03-23
Send etcd version on the wire with connection details

lazypower:
Etcd v2 and v3 have different protocol requirements. It makes sense to
have the version sent along with the connection string so the consumer
application can configure itself appropriately by parsing from the
version identifier if its v2 or v3 specific.

Depends on https://github.com/juju-solutions/interface-etcd/pull/9

### DIFF
--- a/lib/etcdctl.py
+++ b/lib/etcdctl.py
@@ -141,6 +141,18 @@ class EtcdCtl:
         os.environ['ETCDCTL_KEY_FILE'] = key_path
         return check_output(split(command)).decode('ascii')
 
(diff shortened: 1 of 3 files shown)


mbruzek:
+1 to this change, looks safe, but I have not yet tested it. Looks like you have, thanks for adding unit tests.

mbruzek:
I verified the version is on the relationship with deploying a charm built from this branch.
```
juju debug-hooks kubernetes-master/0
(wait for relation hooks)
relation-get
```

PR #82, opened 2017-03-23
Adds version to the action output for a snapshot

lazypower:
etcd snapshots are version specific, and restorations should be performed
on the same etcd version as the snapshot was taken. EG: if you snapshot a
2.3.2 cluster, it is known to fail if you attempt restoration on a 3.0.0
cluster.


mbruzek:
+1 to adding the version to the action!

PR #83, opened 2017-03-23
Cluster string adjustements

lazypower:
Related to https://github.com/kubernetes/kubernetes/issues/43461
partial fix for not all etcd endpoints being transmitted on the
relationship interface.


lazypower:
Closing because the commit history looked really wrong

PR #84, opened 2017-03-23
Cluster string logic fix

lazypower:
Related to kubernetes/kubernetes#43461
partial fix for not all etcd endpoints being transmitted on the
relationship interface.

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -174,6 +174,7 @@ def send_cluster_connection_details(cluster, db):
     key = read_tls_cert('client.key')
     ca = read_tls_cert('ca.crt')
     etcdctl = EtcdCtl()
+    bag = EtcdDatabag()
 
     # Set the key, cert, and ca on the db relation
     db.set_client_credentials(key, cert, ca)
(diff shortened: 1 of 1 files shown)


reactive/etcd.py
```diff
@@ -189,6 +194,7 @@ def send_cluster_connection_details(cluster, db):
 
 @when('db.connected')
 @when('etcd.ssl.placed')
+@when_not('cluster.joined')
```
mbruzek, line 197: The important part is here, this is the inverse of the previous method. I think this is good!

reactive/etcd.py
```diff
@@ -174,13 +174,18 @@ def send_cluster_connection_details(cluster, db):
...
 
     # Set the key, cert, and ca on the db relation
     db.set_client_credentials(key, cert, ca)
 
     port = hookenv.config().get('port')
     # Get all the peers participating in the cluster relation.
     members = cluster.get_peer_addresses()
+    # Append our own address to the membership list, because peers dont self
+    # actualize
+    members.append(bag.private_address)
```
mbruzek, line 187: adding self to the connection string seems like a good idea to me.

mbruzek:
I carefully observed @chuckbutler make and test this change. The before was flipping from 4 addresses (when there was actually 5) to 1 address. After the fix there was 5 addresses all the time. 

+1 from my end!

PR #85, opened 2017-03-24
Resolve issues with cluster scale when removing the leader

lazypower:
This branch removes the erronious logic when tracking the -departed
relation states to remove all participating peers. Instead we are now
relying on the behavior of the -broken hook to only execute on the unit
that is being removed from the cluster. This may not be perfect but
works 100% better than the current implementation.

Resolves https://github.com/kubernetes/kubernetes/issues/43461


marcoceppi (approved):
Ran the tests, checked the code, great addition!

PR #87, opened 2017-03-29
fix etcd command

wwwtyro:


lazypower:
This broke the unit tests on travis :disappointed:   This auto switch basically means we need an update in the unit_tests/tests_etcdctl.py

PR #90, opened 2017-04-10
remove ssl placed state on upgrade

wwwtyro:
don't merge, currently for discussion only

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -109,6 +109,7 @@ def prepare_tls_certificates(tls):
 def remove_states():
     # stale state cleanup (pre rev6)
     remove_state('etcd.tls.secured')
+
     remove_state('etcd.ssl.placed')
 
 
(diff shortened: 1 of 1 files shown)


reactive/etcd.py
```diff
@@ -109,6 +109,7 @@ def prepare_tls_certificates(tls):
 def remove_states():
     # stale state cleanup (pre rev6)
     remove_state('etcd.tls.secured')
+    remove_state('etcd.ssl.placed')
```
Cynerva: This could easily be mistaken as part of the `stale state cleanup` comment above, can we pull them apart? Either with a blank line between, or move it up above or something.

mbruzek:
It looks like travis does not like this, but I am +1 to this change.

lazypower:
Looks like travis blew up when #87 was merged. I don't feel like that is a blocker to this change as its unit-testing the lib files, not the reactive code.

I am +1 to the state + operational changes to the reactive code, as I have run a simple set of validations with @wwwtyro to ensure this is behaving as expected. +1 LGTM

PR #96, opened 2017-05-16
Parse etcd version from etcdctl cli instead of snap info

lazypower:
Snap info changes tab indentation, none of the commands we use to probe
version are accurately retaining any kind of format, and offer no
machine friendly format options to return the version.

This now parses the version info from etcdctl and will make a best
attempt at keeping the version string consistent.

The problem found was that the version output spaces woudl change
depending on snap packages installed. Say you had snap:

ono-mono-topea-banana  - it would spread teh spacing from 3 to 10 in the
output listing. This broke the version parser horribly.


ktsakalozos:
I would prefer a way to get the version in a machine friendly version (yaml, json), but this fix is too critical to block.

+1

PR #110, opened 2018-01-10
Load etcd env variables from bash_aliases

ktsakalozos:
health action is failing and package-client-credentials is grabbing certificates from a hard coded path.

With this patch we grab etcd certificate related env variables from bash_aliases  that holds what we use in layers.yaml options (https://github.com/juju-solutions/layer-etcd/blob/master/reactive/etcd.py#L406)

### DIFF
--- a/actions/health
+++ b/actions/health
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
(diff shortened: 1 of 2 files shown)


Cynerva:
LGTM

PR #111, opened 2018-01-11
Work on snapshot and restore of etcd

ktsakalozos:
The snapshot action by default snapshots v2 data (for backwards compatibility). If you want v3 data to be backed up we need to set keys-version to v3. This gives us backwards compatibility but might not be convenient in the future.

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -19,6 +19,10 @@ snapshot:
       type: string
       default: '/home/ubuntu/etcd-snapshots'
       description: Location to save the etcd snapshot.
+    keys-version:
+      type: string
+      default: 'v2'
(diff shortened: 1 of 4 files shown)


ktsakalozos:
Spoke too soon (?)

ktsakalozos:
Corrected the initialisation process

ktsakalozos:
No, not ready yet. Seems the data-dir changes from etcd 3.0 to etcd 3.1 (/var/snap/etcd/current/etcd0.etcd vs /var/snap/etcd/current)

ktsakalozos:
Tests are in place now. The case where the data-dir is in /var/snap/etcd/current/etcd0.etcd (etcd3 comming from an upgrade) is also handled.

Cynerva:
:+1:

PR #112, opened 2018-01-18
Add network spaces support

Cynerva:
This adds network spaces support to etcd:
* For etcd <-> etcd communication, use `cluster` endpoint's ingress-address instead of private-address
* For etcd <-> client communication, use `db` endpoint's ingress-address instead of private-address
* Add both ingress addresses to certificate SAN list

This should be reviewed and merged together with https://github.com/juju-solutions/interface-etcd/pull/10


hyperbolic2346:
LGTM

PR #113, opened 2018-01-30
Don't set etcd.registered if registration failed

Cynerva:
This fixes a rare deployment problem where one of the units comes up in an "Errored with 0 known peers" state.

The problem occurs when a follower unit's call to `etcdctl register` occurs at the same as a service restart of snap.etcd.etcd on the leader unit. Registration fails, but the unit continues on as if it succeeded, setting `etcd.registered` and never trying again.

The fix is to catch the failure and bail out before setting `etcd.registered` so a retry will occur.


ktsakalozos:
Thank you for this fix! LGTM

PR #115, opened 2018-02-23
Switch to etcd snap to 3.2

ktsakalozos:


Cynerva:
Can we defer this until next release? I'd like to get lots of CI time on it.

PR #119, opened 2018-04-11
Handle unstarted nodes

ktsakalozos:
This PR should address: https://github.com/juju-solutions/bundle-canonical-kubernetes/issues/541

### DIFF
--- a/lib/etcdctl.py
+++ b/lib/etcdctl.py
@@ -95,10 +95,10 @@ class EtcdCtl:
         for unit in raw_member_list:
             if '[unstarted]' in unit:
                 unit_guid = unit.split('[')[0]
+                members['unstarted'] = {'unit_id': unit_guid}
                 if 'peerURLs=' in unit:
                     peer_urls = unit.split(' ')[1].split("=")[-1]
-                members['unstarted'] = {'unit_id': unit_guid,
-                                        'peer_urls': peer_urls}
+                    members['peer_urls'] = peer_urls
                 continue
             unit_guid = unit.split(':')[0]
             unit_name = unit.split(' ')[1].split("=")[-1]
--- a/lib/etcdctl.py
+++ b/lib/etcdctl.py
@@ -93,7 +93,10 @@ class EtcdCtl:
 
         for unit in raw_member_list:
             if '[unstarted]' in unit:
-                members['unstarted'] = {}
+                unit_guid = unit.split('[')[0]
+                peer_urls = unit.split(' ')[1].split("=")[-1]
+                members['unstarted'] = {'unit_id': unit_guid,
+                                        'peer_urls': peer_urls}
                 continue
             unit_guid = unit.split(':')[0]
             unit_name = unit.split(' ')[1].split("=")[-1]
--- a/lib/etcdctl.py
+++ b/lib/etcdctl.py
@@ -98,7 +98,7 @@ class EtcdCtl:
                 members['unstarted'] = {'unit_id': unit_guid}
                 if 'peerURLs=' in unit:
                     peer_urls = unit.split(' ')[1].split("=")[-1]
-                    members['peer_urls'] = peer_urls
+                    members['unstarted']['peer_urls'] = peer_urls
                 continue
             unit_guid = unit.split(':')[0]
             unit_name = unit.split(' ')[1].split("=")[-1]
--- a/lib/etcdctl.py
+++ b/lib/etcdctl.py
@@ -90,11 +90,13 @@ class EtcdCtl:
         # Expect output like this:
         # 4f24ee16c889f6c1: name=etcd20 peerURLs=https://10.113.96.197:2380 clientURLs=https://10.113.96.197:2379  # noqa
         # edc04bb81479d7e8: name=etcd21 peerURLs=https://10.113.96.243:2380 clientURLs=https://10.113.96.243:2379  # noqa
+        # edc0dsa81479d7e8[unstarted]: peerURLs=https://10.113.96.124:2380  # noqa
 
         for unit in raw_member_list:
             if '[unstarted]' in unit:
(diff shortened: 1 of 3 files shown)


lib/etcdctl.py
```diff
@@ -90,10 +90,15 @@ def member_list(self, leader_address=None):
...
         # edc04bb81479d7e8: name=etcd21 peerURLs=https://10.113.96.243:2380 clientURLs=https://10.113.96.243:2379  # noqa
+        # edc0dsa81479d7e8[unstarted]: peerURLs=https://10.113.96.124:2380  # noqa
 
         for unit in raw_member_list:
             if '[unstarted]' in unit:
-                members['unstarted'] = {}
+                unit_guid = unit.split('[')[0]
+                if 'peerURLs=' in unit:
+                    peer_urls = unit.split(' ')[1].split("=")[-1]
+                members['unstarted'] = {'unit_id': unit_guid,
```
hyperbolic2346: What if there are multiple unstarted units? Is it cool to just smash?

lib/etcdctl.py
```diff
@@ -90,10 +90,15 @@ def member_list(self, leader_address=None):
...
+                                        'peer_urls': peer_urls}
```
hyperbolic2346: peer_urls could be undefined here.

lib/etcdctl.py
ktsakalozos: It is not ideal but we are handling them here. If we have two unstarted nodes at the same time (which is twice as rare) this members[unstarted] entry will serve as a stack. The second node will be unregistered and then on the next reactive call we will unregister the first one.

lib/etcdctl.py
ktsakalozos: You are right! Fixed.

lib/etcdctl.py
```diff
@@ -90,10 +90,15 @@ def member_list(self, leader_address=None):
         # Expect output like this:
         # 4f24ee16c889f6c1: name=etcd20 peerURLs=https://10.113.96.197:2380 clientURLs=https://10.113.96.197:2379  # noqa
         # edc04bb81479d7e8: name=etcd21 peerURLs=https://10.113.96.243:2380 clientURLs=https://10.113.96.243:2379  # noqa
+        # edc0dsa81479d7e8[unstarted]: peerURLs=https://10.113.96.124:2380  # noqa
 
         for unit in raw_member_list:
             if '[unstarted]' in unit:
-                members['unstarted'] = {}
+                unit_guid = unit.split('[')[0]
```
wwwtyro, line 97: I'm trying to see if we can use `ETCDCTL_API=3 -w json`, so don't merge this quite yet. ~~It's looking shaky though, so I'll probably retract this.~~

wwwtyro: Okay, this works:

```
$ ETCDCTL_CACERT=/var/snap/etcd/common/ca.crt ETCDCTL_CERT=/var/snap/etcd/common/client.crt ETCDCTL_KEY=/var/snap/etcd/common/client.key ETCDCTL_API=3 etcdctl --endpoints [https://10.205.42.38:2379] -w json member list
{"header":{"cluster_id":9581641538479738059,"member_id":7672478921058476181,"raft_term":2},"members":[{"ID":7672478921058476181,"name":"etcd0","peerURLs":["https://10.205.42.38:2380"],"clientURLs":["https://10.205.42.38:2379"]}]}
```

See https://github.com/coreos/etcd/issues/6915#issuecomment-264043931 for an explanation of why we need the different env vars.

Cynerva: We have to support etcd 2.3 as well. I don't think we can use ETCDCTL_API=3 there, can we?

reactive/etcd.py
```diff
@@ -326,8 +326,21 @@ def register_node_with_leader(cluster):
...
+                # try to self-unregister.
+                try:
+                    leader_address = leader_get('leader_address')
+                    msg = 'Etcd service did not start. Will retry soon.'
+                    etcdctl.unregister(peers[unit]['unit_id'], leader_address)
+                except CalledProcessError:
+                    log('Notice:  Unit failed to unregister', 'WARNING')
+            # we cannot register. State not attainable.
             status_set('waiting', msg)
             return
```
Cynerva, line 345: This looks to me like it will unregister, return, and retry on the next hook (up to 5 min delay). Why not continue on and retry registration immediately?

lib/etcdctl.py
wwwtyro: Good question - I'll try it out.

wwwtyro: Nope, it's no good with the etcdctl we ship with 2.3.8. We'll need to keep the manual string parsing until we can upgrade it.

reactive/etcd.py
ktsakalozos: Indeed this is the intend here, to retry on the next reactive cycle.

I was not able to reproduce the respective issue consistently. It happened only once. So I cannot tell you what would happen if we try to immediately register, I can only mitigate the situation I saw by unregistering the node and waiting.

There are two main reasons why I would not go for the immediately retry.
- Suppose there is something wrong with the node causing it to stay in unstarted state forever. If we see we have an unstarted node  try to unregister and register immediately we will always have a node in an unstarted state. As a result no other node will try to register and our deployment will remain broken. If we give a time window for other nodes to enter the cluster the admin will be able to identify that a certain node is failing and take it down
- I am not sure how long we should wait before we try to register again.

Both of the above points are speculation be cause again this bug is so rare and i cannot reproduce. I would be ok if we decide to not merge this but on the other hand the cluster is already broken at this point so we might as well try to recover be evicting the failing node.

Cynerva: Ah okay, that makes sense. I'm +1 to this then. Thanks for explaining.

PR #124, opened 2018-07-10
support new snapd_refresh from layer-snap

kwmonroe:
Default our etcd charm to use the 'max' possible refresh window,
which is currently to refresh once a month. The leader will get
the initial value and set leader data so all followers (present
and future) will refresh around the same time.

Adding an amulet test for this and fixed up lint issues in the tests.

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -10,6 +10,7 @@ from charms.reactive import when_not
 from charms.reactive import set_state
 from charms.reactive import remove_state
 from charms.reactive import hook
+from charms.reactive.helpers import data_changed
 
 from charms.templating.jinja2 import render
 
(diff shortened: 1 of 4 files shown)


kwmonroe:
WIP until https://bugs.launchpad.net/layer-snap/+bug/1780834 is resolved

reactive/etcd.py
```diff
@@ -414,6 +415,35 @@ def initialize_new_leader():
...
+    if data_changed('etcd_snapd_refresh', timer):
+        log('setting snapd_refresh timer to: {}'.format(timer))
+        leader_set({'snapd_refresh': timer})
+
+
+@when('snap.installed.etcd')
+@when('leadership.changed.snapd_refresh')
+@when_not('leadership.is_leader')
+def set_snapd_timer():
+    ''' Set the snapd refresh.timer on non-leader cluster members. '''
```
hyperbolic2346: Do we need to handle where I'm not the leader, but the leader hasn't set snapd_refresh yet?

kwmonroe: I think we're covered with `@when('leadership.changed.snapd_refresh')`.  If the leader hasn't set it, the leadership.changed flag won't be set.

If you're talking about the case where a follower joins and doesn't know about the leader yet, then yes, layer-snap will set the refresh to a month from yesterday.  But that will quickly be re-set (if it even has a chance to *be* set) as soon as the follower handles the leader hooks.

So, no, i don't think we need a handler for `@when_not(leader)` and `@when_not(leadership.data)`.

kwmonroe:
This is no longer a WIP and is ready to go.  @hyperbolic2346 your question about a follower running the handler before leader data is set got me thinking.  It would have been possible that layer-etcd handlers would run before layer-snap.  The last layer to adjust the refresh wins, so to ensure the leader timer is used across the cluster, we need to make sure layer-etcd does its thing *after* layer-snap is done.

This is handled in the most recent commit by gating snapd_refresh handlers on layer-snap's `snap.refresh.set` flag.

ktsakalozos:
LGTM +1.

PR #125, opened 2018-07-13
Do not limit the restarting of services

### LINKED ISSUE #610: etcd deploy occasionally gets stuck in Errored state on GCE
Observed on two CI runs:
* [test-cdk #909](https://ci.kubernetes.juju.solutions/job/test-cdk/909/) ([logs.tar.gz](https://github.com/juju-solutions/bundle-canonical-kubernetes/files/2189516/logs.tar.gz))
* [test-cdk #943](https://ci.kubernetes.juju.solutions/job/test-cdk/943/) ([logs.tar.gz](https://github.com/juju-solutions/bundle-canonical-kubernetes/files/2189511/logs.tar.gz))

In both cases, ETCD followers are registered to the cluster, but they are unable to establish a connection to the leader, leaving the cluster in an `Errored` state.

Output from an ETCD follower failing to connect to the leader:
```
Jul 12 08:57:32 juju-6fe049-2 etcd[16188]: could not get cluster response from https://252.0.11.1:2380: Get https://252.0.11.1:2380/members: EOF
Jul 12 08:57:32 juju-6fe049-2 etcd[16188]: cannot fetch cluster info from peer urls: could not retrieve cluster information from the given urls
Jul 12 08:57:32 juju-6fe049-2 systemd[1]: snap.etcd.etcd.service: Main process exited, code=exited, status=1/FAILURE
```

I find the `252.0.11.1` address suspicious - that's a FAN IP. Why is it trying to connect over FAN? Is there something about FAN that's causing the connection to fail?

The symptoms are similar to those in https://github.com/juju-solutions/layer-etcd/issues/122 - both occurring because the leader cannot be reached - but I suspect the root cause is different, as the EOF message is absent from ETCD's output in that issue.

ktsakalozos:
Might fix https://github.com/juju-solutions/bundle-canonical-kubernetes/issues/610

### DIFF
--- a/templates/systemd
+++ b/templates/systemd
@@ -18,6 +18,7 @@ ExecStart=/usr/bin/etcd $DAEMON_ARGS
 Restart=on-abnormal
 #RestartSec=10s
 LimitNOFILE=65536
+StartLimitBurst=0
 
 [Install]
 WantedBy=multi-user.target


Cynerva:
LGTM, worth a try

Cynerva:
Wait. @ktsakalozos I just realized we don't even use this systemd template. We use the template generated for snap.etcd.etcd instead.

I'm going to open a follow-up PR to remove this template.

Cynerva:
@ktsakalozos the etcd snap lives here: https://github.com/juju-solutions/etcd-snaps/ - note also the branches for v3.2, v3.1, etc

I do not see an easy way to set this configuration in the snap: https://docs.snapcraft.io/build-snaps/syntax

If we need to do this, we might have no choice but to use a wrapper script for etcd that never terminates.

PR #126, opened 2018-07-13
Remove templates/systemd file that is no longer used

Cynerva:
I'm pretty sure we don't use this template anymore - the snap.etcd.etcd systemd service is managed by the etcd snap, not the charm.

I see no references to this file in the charm, so it should be safe to remove:
```
$ git grep systemd
reactive/etcd.py:    # sorry, some hosts need this. The charm races with systemd and wins.
reactive/etcd.py:    # List of systemd services that will be checked
reactive/etcd.py:    # List of systemd services for which the checks will be removed
```


Cynerva:
@ktsakalozos please take a look at this as a follow-up to https://github.com/juju-solutions/layer-etcd/pull/125. Thanks.

ktsakalozos:
LGTM +1

PR #129, opened 2018-07-31
Fix "Waiting to retry etcd registration" when leader is demoted to follower

Cynerva:
Fixes https://github.com/juju-solutions/layer-etcd/issues/120

Most of the changes in this PR are to clean up `register_node_with_leader` so it is less confusing and more robust.

This also makes it where the leader sets the `etcd.registered` state so that if it ever becomes demoted to a follower, it will not try to re-register.


ktsakalozos:
LGTM +1 lets make get it though the CI

@battlemidget, something to keep an eye on. Thank you.

PR #131, opened 2018-08-01
Get setuptools-scm pinned from the base layer

ktsakalozos:


johnsca (approved):

kwmonroe:
@ktsakalozos as @johnsca noted in irc, I don't see any reason we need path.py in the wheelhouse either.  It came in 2+ years ago here:

https://github.com/juju-solutions/layer-etcd/commit/b07a222fe9c00fd13c0440d1b49ca7ea1143da3e#diff-52f13f84394364d44bc7a2fed930e09dR18

The only ref I can find to `import path` was removed a few months later when the install source code was yanked:

https://github.com/juju-solutions/layer-etcd/commit/dc0f31e5cd798e1a3bd1065e943b6c45f217b151#diff-52f13f84394364d44bc7a2fed930e09dL21

Would you be ok with removing path.py from the wheelhouse.txt as well?

kwmonroe (approved):
LGTM as-in.  Let's try out the removal of path.py in a different PR.

PR #137, opened 2018-08-31
Allow setting specific address where for service to bind to

rmescandon:
snap.etcd.etcd.service can now listen only in unit private or public address, or in any, as so far.

config: added service_network_interface parameter
reactive: react to service_network_interface config changed
templates.etcd2 and templates.etcd3: updated listening addresses

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -20,3 +20,11 @@ options:
       based on the charm deployment date. You may also set a custom string as
       described in the 'refresh.timer' section here:
         https://forum.snapcraft.io/t/system-options/87
+  service_network_interface:
+    type: string
+    default: ""
+    description: |
+      The interface the service binds to. Valid values are 'public' or 'private'
+      to bind to public or private unit address respectively.
+      If an empty or an invalid value is set, the service binds to all network
+      interfaces.
--- a/config.yaml
+++ b/config.yaml
@@ -20,11 +20,9 @@ options:
       based on the charm deployment date. You may also set a custom string as
       described in the 'refresh.timer' section here:
         https://forum.snapcraft.io/t/system-options/87
-  service_network_interface:
-    type: string
-    default: ""
+  bind_to_all_interfaces:
+    type: boolean
+    default: true
     description: |
-      The interface the service binds to. Valid values are 'public' or 'private'
-      to bind to public or private unit address respectively.
-      If an empty or an invalid value is set, the service binds to all network
-      interfaces.
+      The service binds to all network interfaces if true. The service binds
+      only to the first found bind address of each relation if false
--- a/lib/etcd_databag.py
+++ b/lib/etcd_databag.py
@@ -3,11 +3,10 @@ from charmhelpers.core.hookenv import unit_get
 from charmhelpers.core.hookenv import config
 from charmhelpers.core.hookenv import is_leader
 from charmhelpers.core.hookenv import leader_get
-from charmhelpers.core.hookenv import unit_private_ip
-from charmhelpers.core.hookenv import unit_public_ip
 from charmhelpers.core import unitdata
 from charms.reactive import is_state
 from etcd_lib import get_ingress_address
+from etcd_lib import get_bind_address
 
 import string
 import random
@@ -21,10 +20,11 @@ class EtcdDatabag:
     when expanded looks like the following:
 
     {'public_address': '127.0.0.1',
+     'cluster_bind_address': '127.0.0.1',
+     'db_bind_address': '127.0.0.1',
      'cluster_address': '127.0.0.1',
      'db_address': '127.0.0.1',
      'unit_name': 'etcd0',
-     'bind_address': '127.0.0.1',
      'port': '2380',
      'management_port': '2379',
      'ca_certificate': '/etc/ssl/etcd/ca.crt',
@@ -36,7 +36,8 @@ class EtcdDatabag:
 
     def __init__(self):
         self.db = unitdata.kv()
-        self.bind_address = self.get_bind_address()
+        self.cluster_bind_address = self.get_bind_address('cluster')
+        self.db_bind_address = self.get_bind_address('db')
         self.port = config('port')
         self.management_port = config('management_port')
         # Live polled properties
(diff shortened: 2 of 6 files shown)


config.yaml
```diff
@@ -20,3 +20,11 @@ options:
       based on the charm deployment date. You may also set a custom string as
       described in the 'refresh.timer' section here:
         https://forum.snapcraft.io/t/system-options/87
+  service_network_interface:
+    type: string
+    default: ""
+    description: |
+      The interface the service binds to. Valid values are 'public' or 'private'
+      to bind to public or private unit address respectively.
+      If an empty or an invalid value is set, the service binds to all network
```
hyperbolic2346: Are we sure that invalid values binding to all network interfaces is the right call here?

lib/etcd_databag.py
```diff
@@ -94,3 +98,16 @@ def storage_path(self):
...
             return etcd_opts['etcd_data_dir']
+
+    def get_bind_address(self):
+        ''' Returns the address that the service binds to. The value is taken
+        from the config parameter 'service_network_interfaces'.
+        If value has been set to 'public' or 'private', the unit public or
+        private address is considered. Otherwise, service is bound to any
+        available network interface'''
+        service_network_iface = config('service_network_interface')
+        if 'private' in service_network_iface:
```
hyperbolic2346: This will be true if I set it to 'totallybrokenbuthasprivateinsideit'. I would think we would just want to compare the string to private and public.

reactive/etcd.py
```diff
@@ -177,6 +177,17 @@ def follower_config_changed():
     close_open_ports()
 
 
+@when('snap.installed.etcd')
+@when('config.changed.service_network_interface')
+@when_not('etcd.installed')
```
hyperbolic2346: can't change the config after etcd.installed is set?

Cynerva:
Hey, thanks for submitting this. Sorry I haven't had time for a full review yet, but I have a couple quick comments.

We generally avoid using `unit_private_ip` because it assumes there is only a single private network, which is true in many deployments but not all of them. Nowadays we use `network-get` ([docs](https://docs.jujucharms.com/2.4/en/developer-network-primitives)) and the `network_get` charmhelpers equivalent. You can see an example of that here: [link](https://github.com/juju-solutions/layer-etcd/blob/3f652286d5e65571ad48782d4b5fa219d913b9e8/lib/etcd_lib.py#L4-L15). For the work you're doing here, you would need bind-addresses, not ingress-addresses.

Do you have capacity to make this work support multi-network deployments as well? It would be a lot easier for us to merge this if it did.

If you don't have capacity, can you tell us more about why you need this? If it's important, we might be able to merge this work in without multi-network support IF there's a strong disclaimer in the `service_network_interface` description stating that this only works on single-network deployments. Not sure about this - need to see if others on the team are onboard with this stance.

Also, I'm struggling to see the value of the `service_network_interface=public` case - seems like peers (and often, clients) will be connecting over private/internal networks, so I would expect this to fail in any case where the private and public IPs are not the same.

reactive/etcd.py
Cynerva: The `etcd.installed` state is actually an old carryover from pre-snap deployments. Today's deployments do not have it set.

A lot of handlers have `@when_not('etcd.installed')` slapped on them as a way to prevent any handlers from running until the user has migrated to snaps.

There's a note about it here:

https://github.com/juju-solutions/layer-etcd/blob/7307aa988db33687aa8707ae646d833f6bb80886/reactive/etcd.py#L45-L49

hyperbolic2346: now on revision 126 ><

rmescandon:
Hey @Cynerva,

Thanks for such a good information.

The need for this change is starting the service listening only in a private address instead of doing it in any network interface. The approach I'm taking of having a service_network_interface config parameter can be replaced to have a simpler one, like 'bind_to_private_address=true' or 'bind_to_all_interfaces=false' if you see it as a better solution (this way we can prevent setting that to the public interface, what probably makes no sense).
Does it make sense using network_get('db') to obtain the private address? (That information is linked to a specific potential relation). The thing here is that I wouldn't like to expose a way of setting a specific IP but saying to the charm 'please, take a valid private address that you find to bind the service', or at least having such option.

Related to the multi-network deployment. I guess you refer to having relations with peers in different models, don't you?. Does the address where service binds affect it?

Cynerva:
> The approach I'm taking of having a service_network_interface config parameter can be replaced to have a simpler one, like 'bind_to_private_address=true' or 'bind_to_all_interfaces=false' if you see it as a better solution (this way we can prevent setting that to the public interface, what probably makes no sense).

Yeah, that sounds good to me. :+1:

I like `bind_to_all_interfaces=false` since it avoids the "private address" term which could lead to confusion.

> Does it make sense using network_get('db') to obtain the private address?

For the client URL, `network_get('db')` is correct, since clients (kubernetes-master, flannel/calico) connect via the `db` relation.

For the peer URL, we'll need `network_get('cluster')` instead, since peers (other etcd units) connect via the `cluster` relation.

> Related to the multi-network deployment. I guess you refer to having relations with peers in different models, don't you?

Not quite, but good guess. I'm referring to network spaces in Juju, and on MAAS in particular where it's possible to have machines that belong to multiple network spaces. Juju allows operators to bind endpoints to specific network spaces, as a way to signal to the charm that certain types of traffic should be sent over specific networks.

For example, the operator might set something like this in a bundle:
```
etcd:
  charm: cs:~containers/etcd-77
  num_units: 3
  bindings:
    cluster: network-one
    db: network-two
```

to indicate that "cluster" (peer) traffic should be sent over network-one, while "db" (client) traffic is sent over network-two. The charm is responsible for using `network-get` to retrieve the correct addresses from Juju.

Full documentation for running CDK with multiple network spaces is here: https://github.com/juju-solutions/bundle-canonical-kubernetes/wiki/Using-multiple-networks-with-CDK

Cross-model relations are a closely related feature because they also require the use of `network-get`, but we don't currently support them. If I remember right, we would need to make additional changes to the charms for those to work, like passing specific relation IDs to `network-get` and not just endpoint names. In any case, it's not something we need to worry about for now. :)

rmescandon:
Sorry for the delay. The code is ready for another look

hyperbolic2346:
LGTM, but I'd like @Cynerva to take a peek as well.

Cynerva:
LGTM, thanks @rmescandon for seeing this through :+1:

PR #146, opened 2019-01-29
Install `etcd` snap as confined snap

n-pochet:
* According to LP1813678, the installation fails because it tries to
install the `etcd` snap as classic.
* Fixes LP1813678


n-pochet:
Duplicate of #147

PR #148, opened 2019-04-11
Adding patch from cdk to fix snapd_refresh smash by subordinate charm

hyperbolic2346:
Also had to add a check for space only. I wonder if that needs to be ported to cdk. I'll check in the morning...

https://github.com/juju-solutions/kubernetes/pull/212

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ observation.
 We can deploy a single node with the following commands:
 
 ```shell
-juju deploy easyrsa
-juju deploy etcd
+juju deploy ~containers/easyrsa
+juju deploy ~containers/etcd
 juju add-relation etcd easyrsa
 ```
 And add capacity with:
@@ -47,9 +47,7 @@ Juju.
 Health of the cluster can be checked by running a juju action.
 
 ```shell
-juju action do etcd/0 health
-<return response uuid>
-juju action fetch <uuid>
+juju run-action --wait etcd/0 health
 ```
 
 The health is also reported continuously via `juju status`. During initial
@@ -79,7 +77,7 @@ Client certificates can be obtained by running an action on any of the cluster
 members:
 
 ```shell
-juju run-action etcd/0 package-client-credentials
+juju run-action --wait etcd/0 package-client-credentials
 juju scp etcd/0:etcd_credentials.tar.gz etcd_credentials.tar.gz
 ```
 
@@ -111,7 +109,7 @@ cloud provides to us, which can be listed:
 juju list-storage-pools
 ```
 
-#### AWS Storage example
+### AWS Storage example
 
 To add SSD backed EBS storage from AWS, the following example provisions a
 single 10GB SSD EBS instance and attaches it to the etcd/0 unit.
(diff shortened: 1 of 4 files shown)


reactive/etcd.py
```diff
@@ -428,6 +428,12 @@ def process_snapd_timer():
     # Get the current snapd refresh timer; we know layer-snap has set this
     # when the 'snap.refresh.set' flag is present.
     timer = snap.get(snapname='core', key='refresh.timer').decode('utf-8')
+    if not timer or timer.isspace():
```
kwmonroe: `isspace` works here because the timer gets set to an empty string by the subordinate, which layer-snap eventually returns as `\n`:

```
sudo snap set core refresh.timer=''
...
>>> subprocess.check_output(['sudo', 'snap', 'get', 'core', 'refresh.timer']).decode('utf-8')
'\n'
```

How about we check `timer` against the configured value, as in:
```
if not timer or timer != hookenv.config('snapd_refresh')
```

This would mean that etcd would be in charge of the host snap timer, effectively ignoring any subordinate refresh config, but maybe that's ok?

kwmonroe: On second thought, I think it's more important to set the refresh timer to whatever it was before the subordinate smashed it.  Blindly falling back to the charm config may not be what we want.

For example, deploy etcd today, Thursday the 18th.  By default, etcd configures a 'max' refresh timer.  This is translated by layer-snap to "wed3" to signify that the snap should not check for refreshes until the 3rd Wednesday (effectively  a month from yesterday).

If we add a subordinate with no refresh configured to the etcd unit in a couple weeks, it will nullify the core snap refresh timer, and the leader will fall into this method.  With the current patch in place, etcd will set a new timer based on config (still 'max'), which layer-snap will translate into something other than the previously set 'wed3' timer.

All that to say, let's keep checking timer for None or isspace here. If that condition is true, check for previously set timers with `leader_get` and only fall back to config if we don't have a previous value.

hyperbolic2346: Comparing against the config value will catch a situation where the subordinate has a value and the charm does as well. Who should win there? I would vote the main charm.

kwmonroe: Yeah, I agree that the principal charm should always win.

However, since our config defaults to 'max', it's not as simple as comparing the current timer to config since the current timer will never say 'max'.

kwmonroe:
@hyperbolic2346 @johnsca, i've tweaked this according to the previous discussion.  Look ok to you?

Edit: /unhold: ready for review

hyperbolic2346 (commented):
What if timer is not "", but it isn't the value we want? Like they deploy the subordinate with a value for refresh that clashes with the one we want?

kwmonroe:
> What if timer is not "", but it isn't the value we want?

Good point; added docs to note best practice when multiple layer-snap charms are installed on the same machine.

README.md
```diff
@@ -349,16 +360,15 @@ a single node due to the nature of how replicas work in etcd.
...
 are dropped from the snapshot state to enable snapshot restoration. Please
 follow the migration instructions above in the restore action description.
 
-## Contributors
+# Contact
 
-- Charles Butler &lt;[charles.butler@canonical.com](mailto:charles.butler@canonical.com)&gt;
-- Mathew Bruzek  &lt;[mathew.bruzek@canonical.com](mailto:mathew.bruzek@canonical.com)&gt;
+The `etcd` charm is free and open source software created by the
+~containers team at Canonical.
```
hyperbolic2346: No link or mailing list or something? launchpad group?

PR #149, opened 2019-04-17
Making sure cert has proper IP as well as DNS

xtrusia:
In case hookenv.unit_public_ip returns DNS,
TLS cert file has only DNS for servername.
IP is also needed for the other component such as kubernetes-master

Then kubernetes-master faces issue with that.
Because kubernetes-master has IP for it's --etcd-servers option.

There is socket.gethostname already later, I think that
making the result of unit_public_ip real IP makes sense.

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -103,7 +103,6 @@ def missing_relation_notice():
 
 
 @when('certificates.available')
-@when_not('etcd.ssl.placed')
 def prepare_tls_certificates(tls):
     status_set('maintenance', 'Requesting tls certificates.')
     common_name = hookenv.unit_public_ip()
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -108,7 +108,7 @@ def prepare_tls_certificates(tls):
     status_set('maintenance', 'Requesting tls certificates.')
     common_name = hookenv.unit_public_ip()
     sans = set()
-    sans.add(hookenv.unit_public_ip())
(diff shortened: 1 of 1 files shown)


xtrusia:
LP https://bugs.launchpad.net/charm-etcd/+bug/1825110

Cynerva:
I don't think this will fix the issue.

The etcd charm tells the masters to connect to the ingress address of the `db` relation. That happens here: https://github.com/charmed-kubernetes/layer-etcd/blob/f6337557f8be20bc83786cceb5cae2aa66feca77/reactive/etcd.py#L218-L224

The same address is also included in the cert request: https://github.com/charmed-kubernetes/layer-etcd/blob/f6337557f8be20bc83786cceb5cae2aa66feca77/reactive/etcd.py#L112

The problem isn't that the sans list is generated incorrectly, but rather that the cert was not updated at all after the network-get tool started returning IPs instead of hostnames.

I think the way to fix this is to ensure that the `prepare_tls_certificates` handler runs after the network-get output changes. A simple way to do this would be to remove the `@when_not('etcd.ssl.placed')` decorator from the handler. Doing so will cause the handler to run on every Juju hook invocation.

xtrusia:
Hello, Thanks for your description.

Now I understand that.

and I updated commit that just removing etcd.ssl.placed line.

Cynerva:
Thanks for the update. LGTM :+1:

PR #153, opened 2019-05-08
Add core snap resource

Cynerva:
Fixes https://bugs.launchpad.net/charm-kubernetes-master/+bug/1828063 for etcd.

I need to add the core resource to our CI before we can merge this.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -23,6 +23,10 @@ resources:
     type: file
     filename: snapshot.tar.gz
     description: Tarball snapshot of an etcd clusters data.
+  core:
+    type: file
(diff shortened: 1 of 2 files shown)


kwmonroe (approved):
I built and deployed this with both a full and a zero-byte core.snap resource. Both ways worked fine. I thought maybe we'd run into core snap reinstall issues during `juju config etcd channel=foo`; alas, that went fine in both deployments as well.

LGTM.

Cynerva:
Thanks. Merging this to get the initial push to the unpublished channel, then I'll manually attach the zero-byte resource.

PR #156, opened 2019-06-13
Add extra SAN IP entries for each other etcd unit to fix LP:1832568

mskalka:
Uses the cluster relation to grab the db ingress address and add those to the SAN certificate request to mitigate https://bugs.launchpad.net/charm-etcd/+bug/1832568

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -4,6 +4,7 @@ from charms import layer
 
 from charms.layer import snap
 
+from charms.reactive import endpoint_from_flag
 from charms.reactive import when
 from charms.reactive import when_any
 from charms.reactive import when_not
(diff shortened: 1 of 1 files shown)


kwmonroe:
@mskalka I like the spirit of this, but the additional function feels like overkill to me.  Rather than managing extra_sans in the kv and potentially calling `refresh_certificates` and `prepare_tls_certificates` more times than needed, how about we just append the cluster peers inside of the prep function?

I PR'd your PR to show what I mean: https://github.com/mskalka/layer-etcd/pull/1.  LMK how that looks to you.  FWIW, it's working fine in a local test for me -- you can see all 3 peers in my etcd cluster are alts in the server.crt:

```bash
ubuntu@ip-172-31-11-116:/var/snap/etcd/common$ openssl x509 -text -noout -in server.crt | grep -A 1 Alternat
            X509v3 Subject Alternative Name:
                IP Address:172.31.11.116, IP Address:172.31.21.181, IP Address:172.31.9.111, IP Address:252.11.116.1, IP Address:52.53.216.164, DNS:ip-172-31-11-116
```

kwmonroe (approved):

PR #157, opened 2019-06-26
change work directory for restore action to fix LP:1834132

yoshikado:
When running the restore action you will get permission denied since etcd is a confined snap.
https://bugs.launchpad.net/charm-etcd/+bug/1834132
I have changed the working directory for restoring from /var/tmp to /root/tmp.


Cynerva:
LGTM, thanks. I gave this a quick test and confirmed it resolves the issue.

We have a release in progress so I will circle back and merge this once I've confirmed it won't impact the current release.

PR #158, opened 2019-06-27
Handle updates to certs or CA info

johnsca:
If the certificates or CA info is updated (re-issued, transferred to a different CA, etc), the charm needs to handle that by updating the files on disk (already done) and restarting the services (added here).

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -487,6 +487,20 @@ def tls_state_control():
     set_state('etcd.ssl.placed')
 
 
+@when('etcd.ssl.placed')
+@when_any('tls_client.ca.written',
+          'tls_client.server.certificate.written',
+          'tls_client.client.certificate.written')
+def tls_update():
+    ''' Handle changes to the TLS data by ensuring that the service is
+        restarted.
+    '''
+    remove_state('etcd.ssl.placed')
+    remove_state('tls_client.ca.written')
+    remove_state('tls_client.server.certificate.written')
+    remove_state('tls_client.client.certificate.written')
+
+
 @when_not('etcd.pillowmints')
 def render_default_user_ssl_exports():
     ''' Add secure credentials to default user environment configs,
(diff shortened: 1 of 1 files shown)


reactive/etcd.py
```diff
@@ -487,6 +487,20 @@ def tls_state_control():
...
+          'tls_client.server.certificate.written',
+          'tls_client.client.certificate.written')
+def tls_update():
+    ''' Handle changes to the TLS data by ensuring that the service is
+        restarted.
+    '''
+    remove_state('etcd.ssl.placed')
+    remove_state('tls_client.ca.written')
+    remove_state('tls_client.server.certificate.written')
+    remove_state('tls_client.client.certificate.written')
```
Cynerva, line 508: How does this lead to a restart of the etcd service?

johnsca: Hrm.  It seems that it actually doesn't; I was going based on the `upgrade-charm` handler, where it seemed like removing the `etcd.ssl.placed` flag triggers updating the config appropriately, but looking deeper it seems like that doesn't really accomplish much.

One thing it does do is ensure the cert data is re-echoed to the `db` relations, which in practice seems to have been enough for my testing to work where it didn't without this change.  I do agree that it would be better to ensure the config is re-written and the service restarted.

johnsca:
Updated per review.  As per the other PRs (copying here for completeness), this was tested manually on AWS by doing the following:

1) Deploy CDK with the four patched charms & docker subordinate, wait to settle
2) Run some test pod in the cluster (I used busybox and left processes running on the pod)
3) Add Vault to the cluster, wait to settle
4) Unseal Vault, wait to settle
5) Remove EasyRSA, wait to settle
6) Update kubectl config
7) Confirm pod is still functioning

reactive/etcd.py
johnsca: This should be addressed with the latest commit.

Cynerva:
Tested as described in https://github.com/charmed-kubernetes/charm-kubernetes-master/pull/34#issuecomment-510186591

LGTM, merging

PR #159, opened 2019-08-26
Adjust README to specify keys-version

jayk-canonical:
so people following the information here do not lose data


tvansteenburgh (approved):
Thanks!

PR #160, opened 2019-09-20
Fix etcd 2->3 upgrade issue and loss of initial-cluster-token

Cynerva:
Fixes https://bugs.launchpad.net/charm-etcd/+bug/1843497

This fixes two problems with the etcd charm:
1. Any time a new etcd leader unit is elected, the cluster token is lost and a new one is generated in its place.
2. If an upgrade from etcd 2.3 -> 3.x has occurred in the past, then etcd is configured with a different data-dir than usual. This means that if the etcd charm ever renders a new etcd config, then the data-dir changes and all data is lost.

I tested this with two deployments. The first was a direct deployment of CK 1.16 (pre-release from edge) with these changes. I verified that the cluster token was set correctly, the charm did not attempt to move the etcd data, and that the cluster-token was reused when a new leader was elected.

With the second deployment, I started with CK 1.14 running k8s 1.12 and etcd 2.3. From there, I upgraded to etcd 3.0 and k8s 1.13, then followed the upgrade-notes for CK 1.15 (with a new note for upgrading etcd, see \<link to edit in\>). After the upgrade, I verified that all units were in a healthy state, data was not lost in the process, and the cluster-token did not change. I also verified once again that the cluster-token is preserved after a new leader is elected.

### DIFF
--- a/lib/etcd_databag.py
+++ b/lib/etcd_databag.py
@@ -2,7 +2,7 @@ from charms import layer
 from charmhelpers.core.hookenv import unit_get
 from charmhelpers.core.hookenv import config
 from charmhelpers.core.hookenv import is_leader
-from charmhelpers.core.hookenv import leader_get
+from charmhelpers.core.hookenv import leader_get, leader_set
 from charmhelpers.core import unitdata
 from charms.reactive import is_state
 from etcd_lib import get_ingress_address
@@ -70,14 +70,11 @@ class EtcdDatabag:
 
     def cluster_token(self):
         ''' Getter to return the unique cluster token. '''
-        if not is_leader():
-            return leader_get('token')
-
-        if not self.db.get('cluster-token'):
+        token = leader_get('token')
+        if not token and is_leader():
             token = self.id_generator()
-            self.db.set('cluster-token', token)
-            return token
-        return self.db.get('cluster-token')
+            leader_set({'cluster-token': token})
+        return token
 
     def id_generator(self, size=6):
         ''' Return a random 6 character string for use in cluster init.
--- a/lib/etcd_databag.py
+++ b/lib/etcd_databag.py
@@ -73,7 +73,7 @@ class EtcdDatabag:
         token = leader_get('token')
         if not token and is_leader():
             token = self.id_generator()
-            leader_set({'cluster-token': token})
+            leader_set({'token': token})
         return token
 
     def id_generator(self, size=6):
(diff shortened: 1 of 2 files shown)


lib/etcd_databag.py
```diff
@@ -70,14 +70,11 @@ def __init__(self):
 
     def cluster_token(self):
         ''' Getter to return the unique cluster token. '''
-        if not is_leader():
-            return leader_get('token')
-
-        if not self.db.get('cluster-token'):
+        token = leader_get('token')
```
hyperbolic2346, line 73: leader_get('*token*') -> leader_set({'*cluster-token*': token})

Is that correct? It seems like this confusion predates this change.

Cynerva: The heck? Good catch, thanks. I need to figure out how the hell this didn't blow up in testing.

Cynerva: Yep, the followers were not using the correct token:

```
$ juju run --application etcd -- cat /var/snap/etcd/common/etcd.conf.yml '|' grep token
- Stdout: |
    # Initial cluster token for the etcd cluster during bootstrap.
    initial-cluster-token: 'ZYFRQ6'
  UnitId: etcd/0
- Stdout: |
    # Initial cluster token for the etcd cluster during bootstrap.
    initial-cluster-token: 'None'
  UnitId: etcd/1
- Stdout: |
    # Initial cluster token for the etcd cluster during bootstrap.
    initial-cluster-token: 'None'
  UnitId: etcd/2
```

Yet somehow the cluster came up fine anyway. Seems like the tokens aren't actually being used. Weird.

In any case, I fixed the `leader_set` and tested a new deployment. This time, the tokens look good:

```
$ juju run --application etcd -- cat /var/snap/etcd/common/etcd.conf.yml '|' grep token
- Stdout: |
    # Initial cluster token for the etcd cluster during bootstrap.
    initial-cluster-token: 'FU88K7'
  UnitId: etcd/0
- Stdout: |
    # Initial cluster token for the etcd cluster during bootstrap.
    initial-cluster-token: 'FU88K7'
  UnitId: etcd/1
- Stdout: |
    # Initial cluster token for the etcd cluster during bootstrap.
    initial-cluster-token: 'FU88K7'
  UnitId: etcd/2
```

Thanks again for the catch!

hyperbolic2346 (approved):

PR #161, opened 2019-10-04
fix unit tests

kwmonroe:
Fixes https://bugs.launchpad.net/charm-etcd/+bug/1846802. Travis and the unit tests were in sad shape.  Let's fix 'em:

- update travis with only whats needed for unit tests
- remove duplicate leadership layer
- exclude test bits from built artifact
- simpler tox
- fix unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,7 @@
-dist: trusty
-sudo: required
 language: python
 python:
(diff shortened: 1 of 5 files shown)


tvansteenburgh (approved):
Thanks for this.

PR #162, opened 2019-10-22
Fixing issue with restore cluster action

hyperbolic2346:
initial-cluster is set in etcd's config file, but empty. This
resulted in bad data being used for the restore and it would fail.
Added a check to verify the initial-cluster actually contains
data as well as existing.

Fixes #1841484

Tested by snapshotting a 3.0 etcd and restoring on 3.2 and also a 3.2 etcd and restoring on 3.2.

### DIFF
--- a/actions/restore.py
+++ b/actions/restore.py
@@ -92,7 +92,7 @@ def restore_v3_backup():
     check_call(split(cmd))
 
     configfile = open('/var/snap/etcd/common/etcd.conf.yml', "r")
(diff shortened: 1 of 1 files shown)


Cynerva:
LGTM :+1:

kwmonroe:
Cherry picked to `stable` in https://github.com/charmed-kubernetes/layer-etcd/commit/1ea11918e490e04432ea0b9d3d3532e7334efdf4

PR #167, opened 2020-02-06
Fixing execution of `render_default_user_ssl_exports` before etcd is installed

joedborg:
Re. https://bugs.launchpad.net/charm-etcd/+bug/1861483

Have tested this with CK stable and edge.  Have tested an etcd snap upgrade as well as etcd charm upgrade.

Will hold to test with the exact failure case on the bug report.

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -329,7 +329,7 @@ def install_etcd():
     if channel:
         snap.install('etcd', channel=channel, classic=False)
 
-    set_state('etcd.installed')
+    set_state('snap.installed.etcd')
 
 
 @when('snap.installed.etcd')
@@ -526,7 +526,6 @@ def tls_update():
 
     # ensure that certs are re-echoed to the db relations
     remove_state('etcd.ssl.placed')
-
     remove_state('tls_client.ca.written')
     remove_state('tls_client.server.certificate.written')
     remove_state('tls_client.client.certificate.written')
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -8,6 +8,7 @@ from charms.reactive import endpoint_from_flag
 from charms.reactive import when
 from charms.reactive import when_any
 from charms.reactive import when_not
+from charms.reactive import is_state
 from charms.reactive import set_state
 from charms.reactive import remove_state
 from charms.reactive import hook
@@ -149,7 +150,7 @@ def remove_states():
     # stale state cleanup (pre rev6)
     remove_state('etcd.tls.secured')
     remove_state('etcd.ssl.placed')
-    remove_state('etcd.pillowmints')
+    remove_state('etcd.ssl.exported')
 
 
 @when('snap.installed.etcd')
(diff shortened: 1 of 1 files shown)


joedborg:
[Install hook regularly fails due to etcd version check before etcd is installed](https://trello.com/c/k9g6J5zV/1433-install-hook-regularly-fails-due-to-etcd-version-check-before-etcd-is-installed)

reactive/etcd.py
```diff
@@ -308,7 +309,6 @@ def snap_install():
     snap.install('core')
     if channel:
         snap.install('etcd', channel=channel, classic=False)
-    remove_state('etcd.pillowmints')
```
Cynerva, line 311: Why did you remove this?

reactive/etcd.py
```diff
@@ -329,6 +329,8 @@ def install_etcd():
     if channel:
         snap.install('etcd', channel=channel, classic=False)
 
+    set_state('etcd.installed')
```
Cynerva: Setting `etcd.installed` here will confuse the charm and make it think that a migration to snaps needs to occur, won't it?

`etcd.installed` is a legacy state that comes from the old days when we delivered etcd via charm resource, not snap. The charm still uses that state as a signal that the upgrade from pre-snap -> snap has not occurred. See for example this handler:

https://github.com/charmed-kubernetes/layer-etcd/blob/0e450a360598f4e4ac247e343e05b9aba9e1f04f/reactive/etcd.py#L83-L85

reactive/etcd.py
joedborg: I have replaced `etcd.pillowmints` with `etcd.ssl.exported` as that's all it seemed to do anymore.

Cynerva: Right, but this wasn't replaced with `remove_state('etcd.ssl.exported')`, it was just removed. If we don't clear the `etcd.ssl.exported` flag here, then `render_default_user_ssl_exports` won't be re-run after an upgrade from etcd2 -> etcd3 will it?

reactive/etcd.py
```diff
@@ -329,6 +329,8 @@ def install_etcd():
     if channel:
         snap.install('etcd', channel=channel, classic=False)
 
+    set_state('snap.installed.etcd')
```
Cynerva: I don't think this is necessary given that `snap.install('etcd', ...)` sets the same state here: https://github.com/stub42/layer-snap/blob/9f8b9a03f57b528d200d037789db910ca79698b9/lib/charms/layer/snap.py#L58

reactive/etcd.py
joedborg: I think it'll be cleared by L148-149 `@hook('upgrade-charm')` - `def remove_states():`.  I thought this route had to be taken if you were going deb->snap?

reactive/etcd.py
joedborg: Ah, thanks!

reactive/etcd.py
Cynerva: Nah, I'm talking about snap->snap upgrade from etcd 2.3 to 3.x, where the user does this:

```
juju config etcd channel=3.0/stable
```

It'll cause `snap_install` to run again, but with no `upgrade-charm` event.

joedborg: Ah yeah, I see what you mean.  Fixed!  Thanks.

Cynerva:
Thanks. LGTM :+1:

PR #168, opened 2020-02-28
Alarms

barryprice:
Add a Nagios check to alert the operator when etcd alarms are raised

Fixes: https://bugs.launchpad.net/charm-etcd/+bug/1828170


axinojolais:
Can the rationale behind using a file instead of directly querying the alarm list from the check be explained in the code somewhere ?

Thanks !

barryprice:
> Can the rationale behind using a file instead of directly querying the alarm list from the check be explained in the code somewhere ?

I've seen it take a couple of seconds to return on not-particularly-busy clusters, so caching the output seemed to make sense  in order to lower overheads. Added a comment to explain that, but we can discuss other approaches if there are concerns.

tvansteenburgh (approved):

PR #170, opened 2020-04-03
Add support for series upgrade

johnsca:
Pause (disable) the service, set an appropriate status.

Part of https://bugs.launchpad.net/charm-etcd/+bug/1869944

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -19,7 +19,6 @@ from charms.templating.jinja2 import render
 from charmhelpers.core.hookenv import log
 from charmhelpers.core.hookenv import leader_set
 from charmhelpers.core.hookenv import leader_get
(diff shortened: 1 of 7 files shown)


reactive/etcd.py
```diff
@@ -154,10 +155,24 @@ def remove_states():
     remove_state('etcd.ssl.exported')
 
 
+@hook('pre-series-upgrade')
+def pre_series_upgrade():
+    bag = EtcdDatabag()
+    host.service_pause(bag.etcd_daemon)
+    status_set('blocked', 'Series upgrade in progress')
```
Cynerva: Seems like this status could easily get overwritten by other handlers. I recommend setting it in an atexit handler or refactoring the charm to use layer-status.

Cynerva:
LGTM

PR #172, opened 2020-05-01
Update nrpe when upgrading charm

stephenmuss:
I noticed when upgrading the charm that this was missing and that the etcd-alarms check wasn't configured.

Fixes: https://bugs.launchpad.net/charm-etcd/+bug/1876287

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -697,9 +697,13 @@ def initial_nrpe_config(nagios=None):
 
 @when('etcd.installed')
 @when('nrpe-external-master.available')
-@when_not('etcd.nrpe.configured')
 @when_any('config.changed.nagios_context',
           'config.changed.nagios_servicegroups')
+def force_update_nrpe_config():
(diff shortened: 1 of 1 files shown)


tvansteenburgh:
@stephenmuss Thank you!

johnsca:
I'm not clear on how this will cause the NRPE config to be updated, or anything else to be done which wouldn't already have happened anyway?  It looks to me like this would just have the effect of preventing `update_nrpe_config` from running more than once until an `upgrade-charm` hook is triggered, whereas previously, it would be updated whenever the other conditions matched. So, if anything, it would cause the config to be updated *less* frequently rather than more.

Is there something I'm missing?

stephenmuss:
@johnsca I've pushed an additional commit that should address your concerns. Thanks for drawing to my attention.

johnsca (approved):

johnsca:
Thanks!

PR #173, opened 2020-05-28
Fix update_nrpe_config being called too early

johnsca:
The change in PR #172 should have left the affirmative conditions on the `update_nrpe_config` handler but moved them to the `force_update_nrpe_config` handler, leaving the original handler to only be guarded by a negative assertion.  Thus, it is called far too early in the charm lifecycle.

Fixes [lp:1881126](https://bugs.launchpad.net/charmed-kubernetes-testing/+bug/1881126)


Cynerva:
LGTM

PR #179, opened 2020-08-27
Fix v2 API being disabled after upgrade to 3.4

johnsca:
Ensure that the config file is re-rendered when upgrading from the CK 1.16 release of the Etcd charm and then switching the channel to 3.4.

Fixes [lp:1892380](https://bugs.launchpad.net/charm-etcd/+bug/1892380)

### DIFF
--- a/lib/etcd_databag.py
+++ b/lib/etcd_databag.py
@@ -65,7 +65,7 @@ class EtcdDatabag:
         self.server_key = key_path
 
         # Cluster concerns
-        self.cluster = self.db.get('etcd.cluster')
+        self.cluster = self.db.get('etcd.cluster', '')
         self.token = self.cluster_token()
         self.cluster_state = 'existing'
 
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -39,6 +39,7 @@ from etcd_lib import get_ingress_address, get_ingress_addresses
 from shlex import split
 from subprocess import check_call
 from subprocess import check_output
+from subprocess import CalledProcessError
 from shutil import copyfile
 
 import os
@@ -155,6 +156,8 @@ def remove_states():
     remove_state('etcd.ssl.placed')
     remove_state('etcd.ssl.exported')
     remove_state('etcd.nrpe.configured')
+    # force a config re-render in case template changed
+    set_state('etcd.rerender-config')
 
 
 @hook('pre-series-upgrade')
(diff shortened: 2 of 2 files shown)


kwmonroe (approved):
This looks nice.  I esp like the extra logging and better v2/v3 handling in `etcd-version`.  +1.

johnsca:
Testing done:

* Step-wise
  * Steps:
    * Deployed a 3-node cluster at 1.16 (rev 478) with default channel (2.3/stable)
    * Write and read data @ v2
    * Switch channel to 3.0/stable
    * Confirm v2 data can still be read
    * Write and read data @ v3
    * Switch channel to 3.2/stable (highest supported in 1.16 charms)
    * Confirm v2 and v3 data can still be read
    * Upgrade charm to 1.17 (496)
    * Confirm v2 and v3 data can still be read
    * Switch channel to 3.4/stable (highest supported in 1.17 charms)
    * Confirm v2 and v3 data can still be read
    * Upgrade charm to PR
    * Confirm v2 and v3 data can still be read
  * Results:
    * After switch to 3.4/stable, charms report unhealthy and the v2 API is not available, but the v3 API works and the v3 data is available
    * After upgrade to PR, charms report healthy again and both v2 and v3 API and data are available

* "Fast" upgrade
  * Steps:
    * Deployed a 3-node cluster at 1.16 (rev 478) with default channel (2.3/stable)
    * Write and read data @ v2
    * Upgrade charm to PR
    * Confirm v2 data can still be read
    * Switch channel to 3.0/stable
    * Confirm v2 data can still be read
    * Write and read data @ v3
    * Confirm v2 and v3 data can still be read
    * Switch channel to 3.4/stable
    * Confirm v2 and v3 data can still be read
  * Results:
    * Everything looks good at every step

Notes:
  * Env vars changed from v2 to v3:
    * `ETCDCTL_KEY_FILE` -> `ETCDCTL_KEY`
    * `ETCDCTL_CERT_FILE` -> `ETCDCTL_CERT`
    * `ETCDCTL_CA_FILE` -> `ETCDCTL_CACERT`
    * `ETCDCTL_ENDPOINT` -> `ETCDCTL_ENDPOINTS`
  * `etcdctl` 3.4 will complain about the v2 env vars being set when using v3 but otherwise will work fine
  * A "cluster" with a single node can get into a state where the service can't start because, even with a valid value for `initial-cluster`, it expects to be able to talk to at least one peer to come back up again. However, a single node "cluster" is not considered reliable anyway.

PR #180, opened 2020-09-15
Fix initial-cluster-state resetting to "existing" during deployment

Cynerva:
Fixes https://bugs.launchpad.net/charm-etcd/+bug/1867544

This fixes the etcd service on the leader failing to start with
```
cannot fetch cluster info from peer urls: could not retrieve cluster information from the given URLs
```

This happens when the [initialize_new_leader](https://github.com/charmed-kubernetes/layer-etcd/blob/84932ee12bab9d19b733334b7ade25ba0871bb1b/reactive/etcd.py#L446) handler is followed by [tls_update](https://github.com/charmed-kubernetes/layer-etcd/blob/84932ee12bab9d19b733334b7ade25ba0871bb1b/reactive/etcd.py#L550) that re-renders the config with `initial-cluster-state=existing` and restarts the etcd service before etcd has properly bootstrapped.

The fix is to store initial-cluster-state in unitdata so that it is preserved in future re-renders of the config.

### DIFF
--- a/lib/etcd_databag.py
+++ b/lib/etcd_databag.py
@@ -67,12 +67,17 @@ class EtcdDatabag:
         # Cluster concerns
         self.cluster = self.db.get('etcd.cluster', '')
         self.token = self.cluster_token()
-        self.cluster_state = 'existing'
+        self.cluster_state = self.db.get('etcd.cluster-state', 'existing')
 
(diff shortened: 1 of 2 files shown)


joedborg (approved):
Looks good to me

PR #181, opened 2020-12-08
changed etcd snapshot default keys-version to v3

rgildein:
Etcd 3.x has two APIs with two separate databases: etcdv2 and etcdv3.
The latest version of Kubernetes (1.13+) will use the etcdv3 API,
however, the default value of the snapshot action is still using v2,
which use database etcdv2.

Fixes: https://bugs.launchpad.net/charm-etcd/+bug/1840933


Cynerva:
LGTM

PR #182, opened 2021-01-10
Fix package-client-credentials action for new versions

DillonN:
I am quite new to all this, so apoligies if I don't understand the problem exactly

I was having trouble exporting client keys using the `package-client-credentials` action. It was failing with the output:

```
unit-etcd-0:
  UnitId: etcd/0
  id: "6"
  results:
    Stderr: |
      cp: missing destination file operand after 'etcd_credentials/client.crt'
      Try 'cp --help' for more information.
      cp: missing destination file operand after 'etcd_credentials/client.key'
      Try 'cp --help' for more information.
      cp: missing destination file operand after 'etcd_credentials/ca.crt'
      Try 'cp --help' for more information.
  status: completed
  timing:
    completed: 2021-01-09 22:44:26 +0000 UTC
    enqueued: 2021-01-09 22:44:06 +0000 UTC
    started: 2021-01-09 22:44:26 +0000 UTC
```

And the output tarball didn't contain any credentials.

I tracked it down to a discrepency in the environment variable naming. In code there is a version check and current versions use new names: https://github.com/charmed-kubernetes/layer-etcd/blob/fe3911e7e4fcf81a3046e02596cf34159a6a69b2/reactive/etcd.py#L584-L595

But the action script was never updated. I'm not sure how to make it change by version, but I think if it has to be one it should be the current version.

Thanks!

### DIFF
--- a/actions/package-client-credentials
+++ b/actions/package-client-credentials
@@ -5,9 +5,9 @@
 source ~/.bash_aliases
 mkdir -p etcd_credentials
 
-cp $ETCDCTL_CERT_FILE etcd_credentials/client.crt
-cp $ETCDCTL_KEY_FILE etcd_credentials/client.key
-cp $ETCDCTL_CA_FILE etcd_credentials/ca.crt
+cp $ETCDCTL_CERT etcd_credentials/client.crt
+cp $ETCDCTL_KEY etcd_credentials/client.key
+cp $ETCDCTL_CACERT etcd_credentials/ca.crt
 
 # Render a README heredoc
 cat << EOF > etcd_credentials/README.txt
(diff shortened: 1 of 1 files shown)


actions/package-client-credentials
```diff
@@ -5,9 +5,9 @@
 source ~/.bash_aliases
 mkdir -p etcd_credentials
 
-cp $ETCDCTL_CERT_FILE etcd_credentials/client.crt
-cp $ETCDCTL_KEY_FILE etcd_credentials/client.key
-cp $ETCDCTL_CA_FILE etcd_credentials/ca.crt
+cp $ETCDCTL_CERT etcd_credentials/client.crt
+cp $ETCDCTL_KEY etcd_credentials/client.key
+cp $ETCDCTL_CACERT etcd_credentials/ca.crt
```
Cynerva, line 10: Looks like this change is in conflict with https://github.com/charmed-kubernetes/layer-etcd/pull/186. You'll need to resolve the conflict before we can merge this.

Cynerva (changes requested):
It looks like #186 fixed the first part of this already. But it would be nice to get your fix in for the README.txt, too. Can you resolve the merge conflict?

johnsca (approved):
I was able to resolve the conflict and the README.txt changes look good.

PR #183, opened 2021-01-27
Allow restoring etcd cluster

mkalcok:
The `restore` action was previously limited to deployments with single member. Now it can be also performed on the leader unit of etcd cluster.

Caveats:
- Restore action must be performed on the (juju) leader unit
- Restore action is performed only on one member. Other cluster members must sync their data from the member that performed the restore action. Depending on the size of restored data, this can affect host and network load.

LP: https://bugs.launchpad.net/charm-etcd/+bug/1841076/
Doc update: https://github.com/charmed-kubernetes/kubernetes-docs/pull/531


actions/restore.py
```diff
@@ -43,8 +56,11 @@
 
 def preflight_check():
     ''' Check preconditions for data restoration '''
+    if not is_leader():
+        function_fail('This action can be run only on the leader unit')
```
auria: nitpick: "This action can only be run on the leader unit".

auria:
I like how the fix has been cleanly added into the pre-existing layout and helpers.

I only had a minor grammar fix, when the action is run on a unit that is not a leader.

I tested it after deploying the canal bundle and here are a few comments:
* Tested the action on non-leader units and it works as expected (action fails and points to the README, which points to the charm-etcd documentation... which has a section about the restore action and the need to attach the snapshot as a resource)
* Tested the action on the leader unit
  * units are in active/executing state, and report different messages on status (unhealthy with 2 peers, etc.)
  * one of the units went to `waiting/idle` and did not execute the restore process until the `update-status` hook was run
  * in the end, all 3 units, went to `active/idle` and reporting `healthy with 3 peers`

I only wonder if `@hook("cluster-relation-changed") should only set a flag to execute the rejoin handler. The advantage of this approach is that if the hook fails, the unit will be in a state (flag) that would retry to rejoin (no matter which hook is run). Otherwise, it could happen that an operator needs to re-run the restore action to have all units rejoining.

Another suggestion would be to move the units into `maintenance/executing`, and maybe a specific message, when the cluster is running the restore operation.

In any case, I don't see my comments as blockers and it is +1 from me.

Thank you for the fix!

mkalcok:
> I only wonder if `@hook("cluster-relation-changed") should only set a flag to execute the rejoin handler. The advantage of this approach is that if the hook fails, the unit will be in a state (flag) that would retry to rejoin (no matter which hook is run). Otherwise, it could happen that an operator needs to re-run the restore action to have all units rejoining.

I was literary debating these two approaches myself before commit :D but then I looked at the other hooks and saw that they are executing code themselves instead of setting flags so I went this route. However I think your way might be better. One question, would setting the flag in `@hook("cluster-relation-changed")` mean immediate execution of that flag handler?

> one of the units went to waiting/idle and did not execute the restore process until the update-status hook was run

Just for clarification. Did you also check the health/member count manually? Because I noticed that juju status message seem to be often outdated (until next `update-status` run).

auria:
> I was literary debating these two approaches myself before commit :D but then I looked at the other hooks and saw that they are executing code themselves instead of setting flags so I went this route. However I think your way might be better. One question, would setting the flag in `@hook("cluster-relation-changed")` mean immediate execution of that flag handler?

I'm not 100% sure, but I think I've seen this behavior happening (immediate execution). I would suggest marking the code (whenever you feel it) as ready for review to get a second opinion from the kubernetes team about the approach taken. As it has been mentioned before, the kubernetes charms are not at the latest in best practices so it seems to me that using the "set_flag" approach would be closer to the vision for reactive charms. But following similar structure across all charms within a product is also reasonable.

> 
> > one of the units went to waiting/idle and did not execute the restore process until the update-status hook was run
> 
> Just for clarification. Did you also check the health/member count manually? Because I noticed that juju status message seem to be often outdated (until next `update-status` run).

I'm afraid I was only checking `watch -c juju status --color`, so the message may have been outdated although the unit had already restored `etcd`. Would this qualify as a bug, not having the "juju status" description up to date at the end of the hook? If so, I would say it qualifies for a different bug/PR, not this one :)

mkalcok:
Code that sets juju status is [here](https://github.com/charmed-kubernetes/layer-etcd/blob/39154a235f7e6996b94d069938752c98055d0f65/reactive/etcd.py#L94). Is it possible that it got executed before the `cluster-relation-update` hook? Solution here could be to manually execute cluster health check at the end of the rejoin function.

> Tested the action on non-leader units and it works as expected (action fails and points to the README, which points to the charm-etcd documentation... which has a section about the restore action and the need to attach the snapshot as a resource)

Hmm I wouldn't say that this is expected behavior. If the unit is not leader, you should get different error message than if the snapshot is not attached. Could you please try to reproduce this behavior, 'cause it works as expected in my deployment.
```
$ juju run-action --wait etcd/1 restore
unit-etcd-1:
  UnitId: etcd/1
  id: "6"
  message: This action can be run only on the leader unit
  results: {}
  status: failed
  timing:
    completed: 2021-01-28 15:41:41 +0000 UTC
    enqueued: 2021-01-28 15:41:32 +0000 UTC
    started: 2021-01-28 15:41:32 +0000 UTC
```

auria:
> Code that sets juju status is [here](https://github.com/charmed-kubernetes/layer-etcd/blob/39154a235f7e6996b94d069938752c98055d0f65/reactive/etcd.py#L94). Is it possible that it got executed before the `cluster-relation-update` hook? Solution here could be to manually execute cluster health check at the end of the rejoin function.

I have destroyed the test env but it is quite possible. The reactive framework will order the handlers to be run, and I don't think the ordering is deterministic. I will test it again.

> > Tested the action on non-leader units and it works as expected (action fails and points to the README, which points to the charm-etcd documentation... which has a section about the restore action and the need to attach the snapshot as a resource)
> 
> Hmm I wouldn't say that this is expected behavior. If the unit is not leader, you should get different error message than if the snapshot is not attached. Could you please try to reproduce this behavior, 'cause it works as expected in my deployment.

Yes, I mixed both behaviors. Non-leader units fail with a message. If the snapshot resource is not present, units will inform about it. Sorry for the confusion.

mkalcok:
@auria Today I was able to more consistently reproduce the behavior that you described (juju status displaying unhealthy cluster until `update-status` is triggered). It was indeed just a matter of units not running `check_cluster_health()` after other members rejoined.
Last commit has fix to this, I hope I'm not abusing `hookenv.set_relation()` too much here.

actions/restore.py
```diff
@@ -211,10 +212,46 @@ def shasum_file(filepath):
...
+    render('etcd3.conf', conf_path, etcd_conf.__dict__, owner='root',
+           group='root')
+
+
+def rebuild_cluster():
+    """Signal other etcd units to rejoin new cluster."""
+    log('Requesting peer members to rejoin cluster')
+    peer_rel_id = hookenv.peer_relation_id()
+    rejoin_request = uuid4().hex
+    hookenv.relation_set(peer_rel_id, force_rejoin=rejoin_request)
```
Cynerva: Calling relation_set directly is generally not considered good practice. If data is going to be sent over the peer relation, then it should be modeled properly in the relation's interface: [interface-etcd/peers.py](https://github.com/juju-solutions/interface-etcd/blob/master/peers.py)

For what you're trying to do, instead of using the peer relation, I think it might make more sense to use leadership data. Leadership data is an application-wide data store that all units can read, but only the leader can set. If you used that, this would become:

```python
from charmhelpers.core.hookenv import leader_set
...
leader_set(force_rejoin=rejoin_request)
```

reactive/etcd.py
```diff
@@ -602,6 +608,50 @@ def render_default_user_ssl_exports():
...
+            # Bump other members to run check_cluster_health and update status
+            hookenv.relation_set(hookenv.peer_relation_id(),
+                                 cluster_update=uuid4())
+            break
+
+
+@hook('cluster-relation-changed')
+def update_relation(cluster=None):
+    """React to cluster relation change"""
+    force_rejoin_request = hookenv.relation_get('force_rejoin')
```
Cynerva: As mentioned before, I think leadership data would be a better fit for handling the force_rejoin broadcast to followers. If you took that approach, then I think this would become:

```python
@when('leadership.changed.force_rejoin')
@when_not('leadership.is_leader')
def force_rejoin_requested():
    force_rejoin()
    check_cluster_health()
```

There'd be no need to compare this against a value in unitdata.kv() if you already know that the value has changed and that this is a follower, right?

reactive/etcd.py
```diff
@@ -602,6 +608,50 @@ def render_default_user_ssl_exports():
...
+    for _ in range(11):
+        # We need randomized back-off timer because only one unit can be
+        # joining at the same time
+        time.sleep(random.randint(1, 10))
+        register_node_with_leader(None)
+        if is_flag_set('etcd.registered'):
+            log('Successfully rejoined the cluster')
+            # Bump other members to run check_cluster_health and update status
+            hookenv.relation_set(hookenv.peer_relation_id(),
+                                 cluster_update=uuid4())
```
Cynerva: I would remove this relation_set. It will cause a lot of thrashing - if there are 4 followers updating, then every etcd unit will run the entire reactive loop 4 times just to update their status. This kind of thrashing is one of the main reasons that charms like kubernetes-master are as sluggish as they are.

I think it's better to just wait for update-status.

Cynerva (changes requested):

actions/restore.py
johnsca: I think it's also the case that the leader unit wouldn't even get a `cluster-relation-changed` hook for the `force_rejoin` flag, since units don't get notified of updates to their own data.  That's also true, however, for leadership data, and additionally the handlers aren't invoked following an action, so it would depend on another hook invocation for the leader regardless.

My suggestion is to move the `force_rejoin` and `check_cluster_health` functions to `lib/etcd_lib.py` and import and call them directly from the action, as well as importing and calling them from the handlers.  (You'd have to keep a handler for `check_cluster_health` with the same decorators, but it could just call out to the function imported from `etcd_lib`; though you'd have to rename one of them to avoid a conflict.)

reactive/etcd.py
johnsca: To add on to this, the actual reason why you're seeing partial cluster health reports is because as soon as a given unit successfully re-registers itself, it will check the cluster status at that point in time, report it, and then exit the hook.  Given that all units will be running the rebuild / rejoin around the same time, with randomness introduced both by Juju, the cloud environment, and the random sleep, there's plenty of window for a given unit to see the status in a partial state at the point at which it registers itself. Furthermore, in addition to the hook thrashing that George mentioned, there's still a window of opportunity where the units may see partial status, even with the extra hook calls (although the chance of that happening would be lessened due to the extra delay caused by the additional hooks).

I was about to say that it might be worth having the units loop on the cluster health until everything comes back up, but that wouldn't work for the leader because the leader data won't be finalized until the action context ends, so it would be waiting for the other units to respond to a message that it hasn't told Juju it's good to send yet. So I guess the options are either accept the stale status, or do a proper coordinated restart using [the coordinator layer](https://git.launchpad.net/layer-coordinator).  That does get back into the realm of extra hooks, though.

A third option, if you really wanted to fix this properly, would be for the charm to register a service that runs in the background and checks the cluster health, and uses `juju-run` to invoke `update-status` if the health status changes.  This way, it would be responsive regardless of what the hook pattern looked like and would never rely on `update-status` and the 5-minute window.

johnsca: All of that said, in the end I agree with George, and it's probably best to just wait for `update-status` and not worry about it.

It has even been [suggested](https://github.com/canonical/operator/issues/126#issuecomment-581838713) that the unit status really isn't the proper place for reporting info like this anyway and that `active` status really shouldn't have *any* sort of message.  Instead, a proper monitoring stack should be put in place for the cluster, and the unit status would just be used for monitoring the state of the charm itself (with `active` meaning that the unit has successfully done what it's expected to do and isn't blocked or waiting on anything from the user or another charm, hence no message).

actions/restore.py
mkalcok: > I think it's also the case that the leader unit wouldn't even get a cluster-relation-changed hook for the force_rejoin flag

That's not an issue here as the leader unit starts new cluster itself and needs only the other units to rejoin. Other than that, I'll get on the suggested changes.

reactive/etcd.py
mkalcok: I would say that the "thrashing" like this is not the worst thing during the explicit operation like `restore` where high load is to be expected anyway and it's not triggered unexpectedly during normal operation. But since you both seem to agree on this, I'll remove cluster health bumping.

mkalcok:
@johnsca @Cynerva 
I reworked the PR to use leader data instead relation data and removed the forced health check update.

johnsca (approved):
LGTM, thanks for the changes.

Cynerva (approved):

PR #185, opened 2021-02-02
Fix health action

tpsilva:
Health action is currently implemented using a bash script that
queries etcdv2's cluster-health, but it is currently failing
because the API version is not being defined by the script, so
on newer versions of etcd it fails since it defaults to v3.

This patch fixes it by changing the action to use the EtcdCtl
wrapper class.

Fixes: https://bugs.launchpad.net/charm-etcd/+bug/1911437

### DIFF
--- a/actions/actions.py
+++ b/actions/actions.py
@@ -120,12 +120,24 @@ def defrag():
         action_fail_now(e.output)
 
 
+def health():
+    '''Call etcdctl cluster-health
+
+    '''
+    try:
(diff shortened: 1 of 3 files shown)


cjohnston1158:
I tested this by deploying using 3.2/stable, checking health, upgrading to 3.3/stable, checking health, upgrading to 3.4/stable and checking health. The action worked in all scenarios.

johnsca (approved):

PR #164, opened 2019-11-08
Fix change in env vars in etcd3

joedborg:
Fixes: https://bugs.launchpad.net/charm-etcd/+bug/1845389

Tests are currently failing with etcd 3.3.15

PR #186, opened 2021-02-05
Check env vars and use the correct ones for package-client-credentials

cjohnston1158:
etcd version 3.3 introduced a change in env variables that broke the package-client-credentials action. This PR will check to see if the old variables are present (thus <3.3) and use them, or use the new vars if not.

Fixes: https://bugs.launchpad.net/charm-etcd/+bug/1914788


johnsca (approved):
Matches the change in #164; apparently the action was just missed in that PR.

PR #187, opened 2021-03-16
Grafana dashboard

mkalcok:
[LP#1913327](https://bugs.launchpad.net/charm-etcd/+bug/1913327)

To successfully add etcd dashboard to Grafana, following is needed:
- Relate `etcd` and `grafana` applications via `grafana-dashboard` interface.
- Relate `etcd` and `prometheus` applications via `prometheus-manual` interface

Note: `etcd` requires TLS authentication from clients that try to connect to its `/metrics` endpoint. This means that `prometheus` needs client certificate signed by same CA that is trusted by `etcd`. However current `prometheus` charm does not support relation with CA so [new bug](https://bugs.launchpad.net/charm-prometheus2/+bug/1918277) was filed in `prometheus` charm and we need to wait until the [related PR](https://code.launchpad.net/~aluria/charm-prometheus2/+git/prometheus2-charm/+merge/399527) is merged.


config.yaml
```diff
@@ -30,3 +30,14 @@ options:
     description: |
       The service binds to all network interfaces if true. The service binds
       only to the first found bind address of each relation if false
+  prometheus_datasource:
```
auria: I think this config parameter shouldn't exist, but rather infer the "appname" from the existing relation between the prometheus Juju app and the etcd app.

lib/etcd_lib.py
```diff
@@ -1,4 +1,15 @@
-from charmhelpers.core.hookenv import network_get, unit_private_ip
+from charmhelpers.core import hookenv
+from charmhelpers.core.hookenv import config, network_get, unit_private_ip
```
auria: If `hookenv` is imported as is, shouldn't the 3 imported helper funcs be used directly from `hookenv.FUNC`?

lib/etcd_lib.py
```diff
@@ -1,4 +1,15 @@
...
+from charmhelpers.contrib.network.ip import is_address_in_network
+
+from jinja2 import Environment, FileSystemLoader
+from jinja2.exceptions import TemplateNotFound
```
auria: I think this could be removed from the helpers lib and use [1] directly from the reactive script. The only caveat is that your render func uses `<<` and `>>` as variable start/end delimiters, but the default values are used by charmhelpers (`{{` and `}}`). What would you think about using the default delimiters in the template?

1. https://github.com/juju/charm-helpers/blob/master/charmhelpers/contrib/templating/jinja.py#L34

reactive/etcd.py
```diff
@@ -786,6 +800,78 @@ def remove_nrpe_config(nagios=None):
...
+    "endpoint.grafana.joined",
+    "leadership.is_leader"
+)
+@when_not("grafana.configured")
+def register_grafana_dashboard():
+    log("Configuring grafana dashboard", level=hookenv.INFO)
+    grafana = endpoint_from_flag("endpoint.grafana.joined")
+    prometheus = endpoint_from_flag('endpoint.prometheus.joined')
+
+    if not prometheus:
```
auria, line 878: If this possible? `register_grafana_dashboard` handler is only triggered when `prometheus.configured` flag is set, which can only happen if `endpoint.prometheus.joined` is available. Shouldn't we leave `len(prometheus.relations)` fail because `NoneType has no "relations" attribute`? I may be missing more in-depth knowledge about Juju relations, if this could be an issue.

auria (commented):
Change looks good. I've added some suggestions that could be worth reviewing.

I've also tested the change in a test env, and it works as expected:
image

auria:
FWIW, steps to test in a test env:
```
git clone -b grafana-dashboard https://github.com:mkalcok/layer-etcd.git
git clone -b bug/1918277 https://git.launchpad.net/~aluria/charm-prometheus2
cd layer-etcd; make build; cd -
cd charm-prometheus2
# edit src/tox.ini to run with the func tests with --keep-model
# Note: there is a bundle that points to /tmp/charm-builds/etcd
make functional
juju deploy cs:grafana --config install_method=snap --config snap_channel=7/stable
juju add-relation grafana:grafana-source prometheus2
juju add-relation grafana etcd
juju run-action grafana/leader --wait get-admin-password
```

lib/etcd_lib.py
mkalcok: The `<<``>>` delimiters are necessary because the json model contains other "variables" wrapped in `{{``}}` that we want to leave as-is. the only thing that we want to replace is `datasource` so I used different delimiters. And since the charmhelpers function does not support custom delimiters (https://github.com/juju/charm-helpers/issues/367) I had to render the template directly using `jinja2`.

reactive/etcd.py
mkalcok: This is perhaps bit of a paranoid approach. I wasn't sure if it's guaranteed that `@when(endpoint.prometheus.departed)` is executed first, in case that prometheus relation is removed.

lib/etcd_lib.py
mkalcok: Fixed, but I went the other way around and directly imported all the required names from hookenv.

mkalcok:
This PR no longer waits on the prometheus charm (mentioned in the description), it's actually the other way around now. This change needs to be merge before the new prometheus functional test will pass, so I marked this PR as ready for review.

reactive/etcd.py
johnsca: The order between decorated handlers is not guaranteed, but this is a good use-case for `register_trigger` as mentioned above.

(Edit: note to self: "above" can change when you edit comments during a review :smile:)

reactive/etcd.py
```diff
@@ -821,6 +834,78 @@ def remove_nrpe_config(nagios=None):
         nrpe_setup.remove_check(shortname=service)
 
 
+@when('endpoint.prometheus.joined',
+      'leadership.is_leader',
+      'certificates.ca.available')
+def register_prometheus_jobs():
```
johnsca, line 844: I think it might be worth adding a comment calling out that you're explicitly not guarding this with `@when_not('prometheus.configured')` because, while it means that it will be called on every hook once the conditions are met, there isn't really a good way to detect / handle IP address changes without doing that, and as long as the data doesn't change it will be idempotent and shouldn't cause any spurious hooks or anything anyway. (Or you could add an explicit [`is_data_changed()`](https://charmsreactive.readthedocs.io/en/latest/charms.reactive.helpers.html#charms.reactive.helpers.is_data_changed) check to minimize the work done in the hook.)

reactive/etcd.py
```diff
@@ -821,6 +834,78 @@ def remove_nrpe_config(nagios=None):
...
+        level=hookenv.DEBUG)
+    grafana.register_dashboard(name=GRAFANA_DASHBOARD_NAME,
+                               dashboard=dashboard)
+    log('Grafana dashboard "{}" registered.'.format(GRAFANA_DASHBOARD_NAME))
+    set_flag("grafana.configured")
+
+
+@when("endpoint.grafana.departed", "grafana.configured")
+def unregister_grafana_dashboard():
+    clear_flag("grafana.configured")
```
johnsca: Same comment as above.

```suggestion
register_trigger(when_not="endpoint.grafana.joined", clear_flag="grafana.configured")
```

reactive/etcd.py
```diff
@@ -821,6 +834,78 @@ def remove_nrpe_config(nagios=None):
...
+                                    {'targets': targets},
+                                ]
+                            })
+    set_flag('prometheus.configured')
+
+
+@when('endpoint.prometheus.departed')
+def reset_prometheus_config():
+    clear_flag('prometheus.configured')
+    clear_flag('grafana.configured')
```
johnsca: This isn't the flag you want.  It's set whenever any single unit goes away and is never automatically cleared (see [docs](https://charmsreactive.readthedocs.io/en/latest/charms.reactive.relations.html#charms.reactive.endpoints.Endpoint)).  You want `@when_not('endpoint.prometheus.joined')` or even better, use [`register_trigger()`](https://charmsreactive.readthedocs.io/en/latest/charms.reactive.flags.html#charms.reactive.flags.register_trigger) since it's guaranteed to run before decorated handlers (like `register_grafana_dashboard`, below):

```suggestion
register_trigger(when_not="endpoint.prometheus.joined", clear_flag="prometheus.configured")
register_trigger(when_not="endpoint.prometheus.joined", clear_flag="grafana.configured")
```

tox.ini
```diff
@@ -10,8 +10,11 @@ basepython = python3
...
     PYTHONPATH={toxinidir}:{toxinidir}/lib
 deps =
+    jinja2
+    netaddr<=0.7.19
     pyyaml
     pytest
+    pytest-mock
     pytest-cov
     flake8
     git+https://github.com/juju-solutions/charms.unit_test/#egg=charms.unit_test
```
johnsca: This has been released to PyPI so could be updated to:

```suggestion
    charms.unit_test
```

unit_tests/test_etcdctl.py
```diff
@@ -110,6 +118,63 @@ def test_get_connection_string(self):
             'https://1.1.1.1:1111'
         )
 
+    @patch('reactive.etcd.render_grafana_dashboard')
+    @patch('reactive.etcd.set_flag')
```
johnsca: FWIW, [charms.unit_test](https://pypi.org/project/charms.unit-test/) already patches out `set_flag` with a work-alike, so you could also just use `is_flag_set()` to validate and not worry about patching it yourself, but either way works.

wheelhouse.txt
```diff
@@ -2,3 +2,4 @@ charms.templating.jinja2>=1.0.0,<2.0.0
 
 # pin netaddr to avoid pulling importlib-resources from above lib
 netaddr<=0.7.19
+netifaces
```
johnsca: I think this can be dropped?

reactive/etcd.py
```diff
@@ -821,6 +834,78 @@ def remove_nrpe_config(nagios=None):
...
+      'certificates.ca.available')
+def register_prometheus_jobs():
+    log('Registering Prometheus metrics collection.')
+    prometheus = endpoint_from_flag('endpoint.prometheus.joined')
+    cluster = endpoint_from_flag('cluster.joined')
+
+    if not etcd_reachable_from_endpoint(prometheus.endpoint_name):
+        log('Aborting Prometheus metrics collection. Etcd is not reachable by'
+            ' prometheus client', WARNING)
+        return
```
johnsca: This seems like an odd check to include. I guess you're trying to guard against the case where the admin both explicitly told the charm to be selective about which interfaces it listens on and then also connected monitoring via an interface which is not in the set of interfaces it's thus listening on. But you're also not surfacing the error very prominently, so I don't know that it would be more noticeable than just having the scraping fail anyway. And at the end of the day, you still don't *actually* know whether Prometheus can or cannot effectively route traffic to Etcd, since Prometheus could have other routes to the DB ingress address even if different or routing could be broken in some other way even if they appear to be on the same subnet. So I guess I'm wondering if it's worth the extra effort and additional possibility for bugs?

wheelhouse.txt
mkalcok: I don't think so. WIthoutit, I'm getting error in etcd install hook
```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-etcd-2/.venv/lib/python3.8/site-packages/charmhelpers/contrib/network/ip.py", line 39, in <module>
    import netifaces
ModuleNotFoundError: No module named 'netifaces'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-etcd-2/.venv/lib/python3.8/site-packages/charms/reactive/__init__.py", line 71, in main
    bus.discover()
  File "/var/lib/juju/agents/unit-etcd-2/.venv/lib/python3.8/site-packages/charms/reactive/bus.py", line 426, in discover
    _register_handlers_from_file(search_path, filepath)
  File "/var/lib/juju/agents/unit-etcd-2/.venv/lib/python3.8/site-packages/charms/reactive/bus.py", line 466, in _register_handlers_from_file
    _load_module(root, filepath)
  File "/var/lib/juju/agents/unit-etcd-2/.venv/lib/python3.8/site-packages/charms/reactive/bus.py", line 444, in _load_module
    return importlib.import_module(package + module)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/var/lib/juju/agents/unit-etcd-2/charm/reactive/etcd.py", line 43, in <module>
    from etcd_databag import EtcdDatabag
  File "lib/etcd_databag.py", line 8, in <module>
    from etcd_lib import get_ingress_address
  File "lib/etcd_lib.py", line 1, in <module>
    from charmhelpers.contrib.network.ip import is_address_in_network
  File "/var/lib/juju/agents/unit-etcd-2/.venv/lib/python3.8/site-packages/charmhelpers/contrib/network/ip.py", line 46, in <module>
    import netifaces
ModuleNotFoundError: No module named 'netifaces'
```

reactive/etcd.py
mkalcok: You are right here...the check does not definitively prove reachability and it should be removed.

mkalcok: \+ bonus points without that check, `netifaces` are no longer needed in `wheelhouse.txt`

wheelhouse.txt
johnsca: Ah, so it's used in charmhelpers but not listed as an install dependency for that library (I assume since it's part of `contrib` which won't always be used). Fair enough.

unit_tests/test_etcdctl.py
johnsca: On this note, though, charms.unit_test definitely needs better docs.

mkalcok:
Updated PR, including suggestions by @johnsca + one small thing. I'm not sure why but the grafana dashboard tile that displayed number of etcd instances `Up` was showing average over last 100 values resulting in numbers like `2.9` which does not really make sense, so I changed it to `Instant` query which displays only the latest value.

Cynerva (approved):

PR #189, opened 2022-01-13
Fixed etcd charm status on unhealhy state

dparv:
Charms needs to set blocked status if the cluster is unhealthy. This way, using [juju-lint](https://charmhub.io/juju-lint) we will be able to receive alerts based on the blocked state and act proactively on fixing errors.

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -133,7 +133,10 @@ def check_cluster_health():
     bp = "{0} with {1} known peer{2}"
     status_message = bp.format(unit_health, peers, 's' if peers != 1 else '')
 
-    status.active(status_message)
+    if unit_health == "UnHealthy":
+        status.blocked(status_message)
+    else:
+        status.active(status_message)
 
 
 @when('snap.installed.etcd')
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -133,7 +133,7 @@ def check_cluster_health():
     bp = "{0} with {1} known peer{2}"
     status_message = bp.format(unit_health, peers, 's' if peers != 1 else '')
 
-    if unit_health == "UnHealthy":
+    if unit_health in ["UnHealthy", 'Errored']:
         status.blocked(status_message)
     else:
         status.active(status_message)


addyess (approved):
+1. Nothing scary looking here AFAIK.  Wouldn't mind another look though from one more reviewer

reactive/etcd.py
```diff
@@ -133,7 +133,10 @@ def check_cluster_health():
     bp = "{0} with {1} known peer{2}"
     status_message = bp.format(unit_health, peers, 's' if peers != 1 else '')
 
-    status.active(status_message)
+    if unit_health == "UnHealthy":
+        status.blocked(status_message)
+    else:
+        status.active(status_message)
```
addyess: @dparv what about this?  It seems like errored state may also be blocked as well.  it seems like `unit_health` can also be `Errored`

```suggestion
    if unit_health == "Healthy":
        status.active(status_message)
    else:
        status.blocked(status_message)
```

addyess (changes requested):
small nit

reactive/etcd.py
dparv: If the unit_health is Errored then from the looks of the code the charm will raise an exception, which will result in the charm being in error state and will stop the execution further. If you think it might continue executing, we can just do this to keep the change to a minimum:
```
if unit_health == "Healthy":
    status.active(status_message)
else:
    status.blocked(status_message)
```
or
```
if unit_health in ["UnHealthy", 'Errored']:
    status.blocked(status_message)
else:
    status.active(status_message)
```

addyess: oh, well i thought i saw [here](https://github.com/charmed-kubernetes/layer-etcd/pull/189/files#diff-315c141f20f7a537eedea355160e9e523ce48b720b47587e6d41715b9794c19fL127-L131) that the Exception is caught

```python
    # Determine units peer count, and surface 0 by default
    try:
        peers = len(etcdctl.member_list())
    except Exception:
        unit_health = "Errored"
        peers = 0

    bp = "{0} with {1} known peer{2}"
    status_message = bp.format(unit_health, peers, 's' if peers != 1 else '')

    status.active(status_message)
```

am i missing something?

dparv: Hm, you are right, but still this looks better imho

```python
if unit_health in ["UnHealthy", 'Errored']:
    status.blocked(status_message)
else:
    status.active(status_message)
```

addyess: @dparv -- sweet.  go for it!

PR #194, opened 2022-02-09
snap connect removable-media when storage is attached

hemanthnakkina-zz:
removable-media interface is not of type autoconnect and
so snap connect removable-media should be done explicity.
In case of data-volume-attached and etcd is installed via
snap, ensure snap connect is executed for removable-media
interface.

Resolves: LP#1960393

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -907,6 +907,13 @@ def register_grafana_dashboard():
     set_flag("grafana.configured")
 
 
+@when("snap.installed.etcd", "data.volume.attached")
+@when_not("snap.connected.removable_media")
+def snap_connect_external_storage():
+    snap.connect("etcd:removable-media", ":removable-media")
+    set_flag("snap.connected.removable_media")
+
+
 def volume_is_mounted(volume):
     """Takes a hardware path and returns true/false if it is mounted"""
     cmd = ["df", "-t", "ext4"]


reactive/etcd.py
```diff
@@ -907,6 +907,13 @@ def register_grafana_dashboard():
     set_flag("grafana.configured")
 
 
+@when("snap.installed.etcd", "data.volume.attached")
+@when_not("snap.connected.removable_media")
```
addyess, line 911: is there an analog case to be considered??

```python
@when("snap.installed.etcd", "snap.connected.removable_media")
@when_not("data.volume.attached")
def snap_disconnect_external_storage():
    snap.disconnect("etcd:removable-media")
    clear_flag("snap.connected.removable_media")
```


sadly though, i don't see `disconnect` available anywhere in the layer.  so maybe i'm just fooling myself. 

https://git.launchpad.net/layer-snap/tree/lib/charms/layer/snap.py


I'm ok with this change, is there any testing we could add?  unit/integration?

addyess (commented):
comments about added testing

reactive/etcd.py
hemanthnakkina-zz: Hi Adam

Thanks for the review.

I haven't added unit test cases as the new function contain only 2 lines of code and asserting if the 2 lines are hit does not make sense to me especially with no conditional operations.
Looked into the integration tests, I can see a way to add storage to the charm in amulet tests [1] but not sure who is supposed to create corresponding storage-pool as I do not see any such definitions in amulet module. 
Also I do not see any CI checks running against the PR. Can you clarify on this to proceed with integration tests. 

[1] https://github.com/juju/amulet/blob/master/amulet/deployer.py#L167-L174

addyess: You're right.  in that it's somewhat silly to add unit tests for two lines.  Integration tests would be nice, but now i see this repo isn't setup to run integration on pull requests -- and the fact that it's using amulet states it's out of date a bit as most of the other kubernetes parts are using `pytest_operator` for integration tests.

i don't see ANYTHING that really makes this worse, and if you've confirmed this addresses your issue it seems really safe to include.

I'll run the integration tests with this change locally to confirm

addyess: Well, running the integration tests was a bust -- but i was able to deploy safely with your changes, and though i haven't exercised this new block -- nothing else is broken.

PR #196, opened 2022-04-21
Cleaned-up tests, converted integration tests to use pytest operator and various bug fixes

berkayoz:
Most of the fixes were related to how `etcdctl` now requires `ETCDCTL_API=2` for using older commands that utilize the v2 API. Updating these to use the v3 api might be a good idea after this.

Some actions also fail due to cluster not being settled, added retry loops to address this issue.

The `cluster-relation-broken` hook is replaced by the `stop` hook. I believe it is related to a Juju change. This fixes [launchpad#1967569](https://bugs.launchpad.net/charm-etcd/+bug/1967569).

Removing the etcd member from cluster may fail at times due to cluster not being settled yet. Implemented a random back off timer for removal of the departing unit to give some time for cluster to settle. The hook currently fails if the member can not be removed and would require intervention. Dangling members aren't probably desired and would be harder to correct in a deployment.

The `restore` action uses `is_v3_backup()` to determine the snapshot's etcd version. Updated the wildcard search to use `*/wal` instead of `*/db` since both versions can contain a db file now. `*/wal` only exists for v2.

Converted the integration tests to use pytest-operator and updated tests to be more precise with the assertions. 

Cleaned up the tests and updated `tox.ini`.


lib/etcdctl.py
```diff
@@ -90,12 +91,14 @@ def member_list(self, leader_address=False):
...
             peer_urls = unit.split(" ")[2].split("=")[-1]
             client_urls = unit.split(" ")[3].split("=")[-1]
+            is_leader = unit.split(" ")[4].split("=")[-1]
 
             members[unit_name] = {
                 "unit_id": unit_guid,
                 "name": unit_name,
                 "peer_urls": peer_urls,
                 "client_urls": client_urls,
+                "is_leader": True if is_leader == "true" else False,
```
neoaggelos:
```suggestion
                "is_leader": is_leader == "true",
```

neoaggelos (commented):
Overall this looks OK to me, but leaving it up to the CK team to approve.

actions/restore.py
```diff
@@ -190,11 +191,20 @@ def probe_forked_etcd():
...
+    loop = 0
+    MAX_WAIT = 10
+
+    while loop < MAX_WAIT:
+        try:
+            cmd = "/snap/bin/etcd.etcdctl member list"
+            members = check_output(split(cmd))
+            member_id = members.split(b',')[0].decode('utf-8')
+            break
+        except:
```
lferran: instead of catching all exceptions, It would be better to catch only those exceptions that you care about

actions/restore.py
```diff
@@ -227,7 +237,13 @@ def dismantle_cluster():
     for name, data in etcdctl.member_list(endpoint).items():
         if name != my_name:
             log('Disconnecting {}'.format(name), hookenv.DEBUG)
-            etcdctl.unregister(data['unit_id'], endpoint)
+            for _ in range(5):
+                try:
+                    etcdctl.unregister(data['unit_id'], endpoint)
+                    break
+                except:
```
lferran: same as above, it's better to be specific about which exceptions are handled. This way if something unexpected happens will not be swallowed by the try catch.

actions/restore.py
```diff
@@ -91,12 +92,12 @@ def unpack_resource():
 
 def is_v3_backup():
     ''' See if the backup file contains a db file indicating a v3 backup '''
-    cmd = "tar -tvf {0} --wildcards '*/db'".format(SNAPSHOT_ARCHIVE)
+    cmd = "tar -tvf {0} --wildcards '*/wal'".format(SNAPSHOT_ARCHIVE)
```
lferran, line 97: I think the docstring of the function should be updated. Or else adding a comment would be helpful. Similar to what you mention in the PR description.

actions/snap-upgrade.py
```diff
@@ -159,6 +159,8 @@ def has_migrated_from_deb():
     backup_package = action_get('target')
     backup = action_get('backup')
     channel = config('channel')
+    if channel == "auto":
+        channel = "3.4/stable"
```
lferran: why is that? maybe it needs a comment?

berkayoz: Leaving the channel on `auto` lets the charm determine which channel to use, I basically mirrored this [function](https://github.com/charmed-kubernetes/layer-etcd/blob/a036f794faef093e6eb173b88fb7fc68444c6952/reactive/etcd.py#L88). The problem is the snap-upgrade action didn't have this default fallback and tried to refresh into channel `auto` which does not exist. Adding a comment is a good idea though.

actions/restore.py
berkayoz: Sure thing, missed that part. Will update it.

actions/restore.py
```diff
@@ -90,13 +91,13 @@ def unpack_resource():
...
-    cmd = "tar -tvf {0} --wildcards '*/db'".format(SNAPSHOT_ARCHIVE)
+    ''' See if the backup file does not contain a wal file indicating a v3 backup '''
+    cmd = "tar -tvf {0} --wildcards '*/wal'".format(SNAPSHOT_ARCHIVE)
     try:
         check_call(split(cmd))
     except CalledProcessError:
-        return False
-    return True
+        return True
+    return False
```
addyess, line 102: i wish i had more context here as to why there's a new command and flipping return values.  Is it because only v3 fails on `/wal` so 

if this command succeeds:
* v1 or v2
if it fails:
* v3

actions/restore.py
```diff
@@ -190,11 +191,20 @@ def probe_forked_etcd():
...
 
-    raw_update = "/snap/bin/etcd.etcdctl member update {0} http://{1}:{2}"
+    loop = 0
+    MAX_WAIT = 10
+
+    while loop < MAX_WAIT:
+        try:
+            cmd = "/snap/bin/etcd.etcdctl member list"
+            members = check_output(split(cmd))
+            member_id = members.split(b',')[0].decode('utf-8')
```
addyess: Why do we now split on comma (`b","`) rather than colon (`b':'`)?  

Has the output of this shell command changed and we need to support two different output formats?

actions/restore.py
```diff
@@ -190,11 +191,20 @@ def probe_forked_etcd():
 def reconfigure_client_advertise():
     ''' Reconfigure the backup to use host network addresses for client advertise
         instead of the assumed localhost addressing '''
-    cmd = "/snap/bin/etcd.etcdctl member list"
-    members = check_output(split(cmd))
-    member_id = members.split(b':')[0].decode('utf-8')
 
-    raw_update = "/snap/bin/etcd.etcdctl member update {0} http://{1}:{2}"
+    loop = 0
+    MAX_WAIT = 10
```
addyess, line 198: `MAX_WAIT` really is "MAX_LOOPS`.  I don't see any "waiting" going on (ie -- sleeps or whatever).  Did you have an intent to wait?

actions/restore.py
```diff
@@ -190,11 +191,20 @@ def probe_forked_etcd():
...
+
+    while loop < MAX_WAIT:
+        try:
+            cmd = "/snap/bin/etcd.etcdctl member list"
+            members = check_output(split(cmd))
+            member_id = members.split(b',')[0].decode('utf-8')
+            break
+        except CalledProcessError:
+            loop = loop + 1
```
addyess, line 207: could be nice to log here that we're going for another loop and perhaps even wait for a bit before retrying

actions/restore.py
```diff
@@ -227,7 +237,13 @@ def dismantle_cluster():
...
         if name != my_name:
             log('Disconnecting {}'.format(name), hookenv.DEBUG)
-            etcdctl.unregister(data['unit_id'], endpoint)
+            for _ in range(5):
+                try:
+                    etcdctl.unregister(data['unit_id'], endpoint)
+                    break
+                except EtcdCtl.CommandFailed:
+                    # Randomized back-off timer to let cluster settle
+                    time.sleep(random.randint(1, 3))
```
addyess: i like the idea of backoff more than i like random here.  If you want to 
sleep based on range loop variable rather than `randint`, that could work out better

```suggestion
                    time.sleep(_+1)  
```

actions/snap-upgrade.py
addyess: why not call that function?

lib/etcdctl.py
```diff
@@ -90,12 +91,14 @@ def member_list(self, leader_address=False):
             unit_name = unit.split(" ")[1].split("=")[-1]
             peer_urls = unit.split(" ")[2].split("=")[-1]
             client_urls = unit.split(" ")[3].split("=")[-1]
+            is_leader = unit.split(" ")[4].split("=")[-1]
```
addyess: the above sample text doesn't show this "leader" field.  Has the output changed from a different etcd version?  Does this still support the prior etcd versions?

lib/etcdctl.py
```diff
@@ -178,7 +181,9 @@ def run(self, arguments, endpoints=None, api=3):
                 command.insert(2, endpoints)
 
         try:
-            return check_output(command, env=env).decode("utf-8")
+            return sp_run(
+                command, env=env, capture_output=True, check=True
+            ).stdout.decode("utf-8")
```
addyess: why does this need to change?  i don't understand what's different.

tests/integration/test_actions.py
```diff
@@ -0,0 +1,123 @@
...
+    "ETCDCTL_KEY=/var/snap/etcd/common/client.key",
+    "ETCDCTL_CERT=/var/snap/etcd/common/client.crt",
+    "ETCDCTL_CACERT=/var/snap/etcd/common/ca.crt",
+]
+
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
+    charm = await ops_test.build_charm(".")
+    await ops_test.model.deploy("cs:~containers/easyrsa-441")
```
addyess: lets use a modern charm version:

```suggestion
    await ops_test.model.deploy("easyrsa")
```

tests/integration/test_actions.py
```diff
@@ -0,0 +1,123 @@
...
+    resources = {"snapshot": "snapshot.tar.gz"}
+    await ops_test.model.deploy(charm, resources=resources)
+    await ops_test.model.add_relation("easyrsa:client", "etcd:certificates")
+    await ops_test.model.wait_for_idle(wait_for_active=True, timeout=60 * 60)
```
addyess: it shouldn't take an hour for this deployment:

```suggestion
    await ops_test.model.wait_for_idle(wait_for_active=True, timeout=5 * 60)
```

tests/integration/test_actions.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+async def test_snapshot_restore(ops_test: OpsTest):
```
addyess: i keep noticing this:

```python
ops_test.model.applications["etcd"].units[0]
```

lets use a pytest fixture:

```python
@pytest.fixture
def etcd_unit(ops_test: OpsTest):
    yield ops_test.model.applications["etcd"].units[0]

async def test_snapshot_restore(ops_test: OpsTest, etcd_unit: Unit):
   ...
   action = await etcd_unit.run_action(...)
...

File: tests/integration/test_actions.py [hunk 16]
```diff
@@ -0,0 +1,123 @@
...
+
+    await delete_data(ops_test)
+    assert not await is_data_present(ops_test, "v2")
+    assert not await is_data_present(ops_test, "v3")
+
+    # Below code is better but waiting for python-libjuju #654 fix, can't attach binary files yet due to the bug
+    # with open(filenames["v2"], mode='rb') as file:
+    #     ops_test.model.applications["etcd"].attach_resource("snapshot", filenames["v2"], file)
+
+    await ops_test.model.wait_for_idle(wait_for_active=True, timeout=60 * 60)
```
addyess: 5-min is probably fine... 60's a bit long

tests/integration/test_actions.py
```diff
@@ -0,0 +1,123 @@
...
+    await action.wait()
+    assert action.status == "completed"
+    assert await is_data_present(ops_test, "v2")
+    assert not await is_data_present(ops_test, "v3")
+
+    # Below code is better but waiting for python-libjuju #654 fix, can't attach binary files yet due to the bug
+    # with open(filenames["v3"], mode='rb') as file:
+    #     ops_test.model.applications["etcd"].attach_resource("snapshot", filenames["v3"], file)
+
+    await ops_test.model.wait_for_idle(wait_for_active=True, timeout=60 * 60)
```
addyess: 5min

tests/integration/test_deb_snap_migrate.py
```diff
@@ -0,0 +1,76 @@
+import pytest
```
addyess: having multiple integration tests python modules will mean your constantly creating machines, and tearing them down, building new models over and over.  If that's the intent -- okay, but it's time consuming.  

Consider trying to get as much done in one model as possible

tests/integration/test_deb_snap_migrate.py
```diff
@@ -0,0 +1,76 @@
...
+    "ETCDCTL_CERT_FILE=/var/snap/etcd/common/client.crt",
+    "ETCDCTL_CA_FILE=/var/snap/etcd/common/ca.crt",
+    "ETCDCTL_KEY=/var/snap/etcd/common/client.key",
+    "ETCDCTL_CERT=/var/snap/etcd/common/client.crt",
+    "ETCDCTL_CACERT=/var/snap/etcd/common/ca.crt",
+]
+
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
```
addyess: this looks like the same setup as the previous module?  Do we really need a new juju model and two new machines etc?

tests/integration/test_deb_snap_migrate.py
```diff
@@ -0,0 +1,76 @@
...
+    charm = await ops_test.build_charm(".")
+    await ops_test.model.deploy("cs:~containers/easyrsa-441")
```
addyess: please use `easyrsa`

tests/integration/test_deb_snap_migrate.py
```diff
@@ -0,0 +1,76 @@
...
+    await ops_test.model.deploy("cs:~containers/etcd-655")
```
addyess: wait, you just built the etcd charm -- but your deploying from the charmstore?  did you mean:
```suggestion
    await ops_test.model.deploy(charm)
```

tests/integration/test_deploy.py
```diff
@@ -0,0 +1,106 @@
...
+    "ETCDCTL_CERT_FILE=/var/snap/etcd/common/client.crt",
+    "ETCDCTL_CA_FILE=/var/snap/etcd/common/ca.crt",
+    "ETCDCTL_KEY=/var/snap/etcd/common/client.key",
+    "ETCDCTL_CERT=/var/snap/etcd/common/client.crt",
+    "ETCDCTL_CACERT=/var/snap/etcd/common/ca.crt",
+]
+
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
```
addyess: third model?  5th and 6th machine... how long does this take?

tests/integration/test_deploy.py
```diff
@@ -0,0 +1,106 @@
...
+
+async def _get_leader(units: List[Unit]) -> Unit:
+    for unit in units:
+        if await unit.is_leader_from_status():
+            return unit
+
+
+async def test_leader(ops_test: OpsTest):
+    leader = await _get_leader(ops_test.model.applications["etcd"].units)
+    assert leader is not None
```
addyess: this is really a test of juju, not of etcd.  One of them is going to be the leader whether it likes it or not -- and you've only deployed one unit -- so it's gonna be the leader.  I think this is fruitless and probably not a great test

tests/integration/test_deploy.py
```diff
@@ -0,0 +1,106 @@
...
+    charm = await ops_test.build_charm(".")
+    await ops_test.model.deploy("cs:~containers/easyrsa-441")
+    resources = {"snapshot": "snapshot.tar.gz"}
+    await ops_test.model.deploy(charm, resources=resources)
+    await ops_test.model.add_relation("easyrsa:client", "etcd:certificates")
+    await ops_test.model.wait_for_idle(wait_for_active=True, timeout=60 * 60)
+
+
+async def test_status(ops_test):
+    assert ops_test.model.applications["etcd"].units[0].workload_status == "active"
```
addyess: it has to be active, the previous test guarantees it.

tests/unit/test_etcdctl.py
```diff
@@ -57,23 +57,25 @@ def test_unregister(self, etcdctl):
 
     def test_member_list(self, etcdctl):
         with patch("etcdctl.EtcdCtl.run") as comock:
-            comock.return_value = "7dc8404daa2b8ca0: name=etcd22 peerURLs=https://10.113.96.220:2380 clientURLs=https://10.113.96.220:2379\n"  # noqa
+            comock.return_value = "7dc8404daa2b8ca0: name=etcd22 peerURLs=https://10.113.96.220:2380 clientURLs=https://10.113.96.220:2379 isLeader=true\n"  # noqa
```
addyess: I get the feeling that newer versions of the command return this new `isLeader` flag.  Your charm needs to span multiple snap versions, not just the most recent.  Be sure to account for the old output

addyess (changes requested):
Whew... this is a BIG PR.  I like the effort here.  Hopefully i didn't burden you with way too many questions.  

overall,
* try to support multiple revs of the etcd snap since the charm is configurable to do so.
* don't split up the integration tests into different modules unless you REALLY REALLY REALLY want to keep building the charm over and over, and deploying into new models over and over.  This is really time consuming. 
*

tox.ini
```diff
@@ -3,29 +3,45 @@ max-line-length = 120
...
         --cov lib --cov-fail-under 33 \
-        --tb native -s {posargs}
+        --tb native -s {posargs} {toxinidir}/tests/unit
+
+[testenv:integration]
+deps =
+    pytest
+    pytest-operator
+    charms.unit_test
+commands = pytest --tb native --show-capture=no --disable-pytest-warnings --log-cli-level=INFO -s {posargs} {toxinidir}/tests/integration
```
addyess, line 38: OH!  this needs to be added to the .github actions. 

see 
https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/main/.github/workflows/main.yml#L34-L53

actions/restore.py
lferran: What about splitting the retry or retry & wait logic in a separate function/decorator? It will be easier to unit test too.

There are libraries that offer this kind of utils, like [retry](https://pypi.org/project/retry/) or [backoff](https://pypi.org/project/backoff/).

addyess: or [tenacity](https://tenacity.readthedocs.io/en/latest/)

the trouble with libraries(like these, but not necessarily) is making sure they have good back support for releases like xenial (yes, we still support python 3.6)

actions/restore.py
berkayoz: I'm not exactly sure but I think `tar -tvf {0} --wildcards '*/db'` command returned successfully for both v1/v2 and v3 since both contained a `db` file. There were 2 approaches we could take to determine the version, which is either checking the directory structure(db file is at root of the archive for v3, for v2 it is in some other folder) or checking if a `wal` file exists. I've went ahead with the `wal` option, to me which seems more solid and future proof. Will add a comment describing the approach. We can go with something else if there are better suggestions.

actions/restore.py
berkayoz: I believe this might be an oversight on my side, the `etcdctl` on v3 api uses commas instead of colons in the member list output. Tried utilizing the v2 API with `ETCDCTL_API=2` but couldn't get it working on the `v3` version. Will try a version detection/fallback to old api approach for this.

actions/restore.py
berkayoz: Seems like I've forgot to add `sleep`, I aim to retry the command a couple of times until etcd cluster settles. I will be fixing it.

tests/integration/test_deb_snap_migrate.py
berkayoz: This is basically a mirror of the old tests, the local build is used for charm refresh. I believe it is a kind of test, but not sure if it's a great test.

tests/integration/test_deploy.py
berkayoz: It takes a bit longer than a single module but, not too long. But I agree that we should go with a single module/deployment, will combine these into one.

tests/unit/test_etcdctl.py
berkayoz: Again another oversight on my side, we can remove this since there are no uses for it yet.

actions/restore.py
berkayoz: Found the reason why `ETCDCTL_API=2` wasn't working, adding the `--enable-v2` flag was needed on the etcd fork while restoring. Will be going with this fix.

tests/integration/test_deb_snap_migrate.py
addyess: i -- dunno but this looks wrong.  There's no need to deploy a FIXED version of the charm `cs:~containers/etcd-655` rather than the new charm we're trying to test in a PR.  Also we should deploy from `easyrsa` from charmhub with

```python
    await ops_test.model.deploy("easyrsa")
```

.github/workflows/tox.yaml
```diff
@@ -37,3 +37,24 @@ jobs:
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          provider: vsphere
+          credentials-yaml: ${{ secrets.CREDENTIALS_YAML }}
+          clouds-yaml: ${{ secrets.CLOUDS_YAML }}
+          bootstrap-options: "--model-default datastore=vsanDatastore --model-default primary-network=VLAN_2763"
```
addyess: could i offer this -- it will speed up the integration tests run by using an image for the controller that's already got the VM better configured for juju in focal

```suggestion
          clouds-yaml: ${{ secrets.CLOUDS_YAML }}
          bootstrap-constraints: "arch=amd64 cores=2 mem=4G"
          bootstrap-options: "${{ secrets.FOCAL_BOOTSTRAP_OPTIONS }} --model-default datastore=vsanDatastore --model-default primary-network=VLAN_2763"         
```

actions/snap-upgrade.py
```diff
@@ -160,6 +162,11 @@ def has_migrated_from_deb():
     backup = action_get('backup')
     channel = config('channel')
 
+    # We need to determine a default fallback channel since, there isn't an "auto" channel
+    channel = get_target_etcd_channel()
```
addyess: You can remove this line right?

```suggestion
    # We need to determine a default fallback channel since, there isn't an "auto" channel
    channel = get_target_etcd_channel()
```

lib/etcdctl.py
addyess: what's this for

addyess (changes requested):
still have a few questions

actions/snap-upgrade.py
berkayoz: `channel` gets passed to the snap command as the refresh channel. The charm has an "auto" setting for the channel, and the appropriate channel is determined by `get_target_etcd_channel()`. Not having this breaks the snap command since there isn't an `auto` channel for etcd.

lib/etcdctl.py
berkayoz: As far as I understand `check_output` doesn't capture `stderr`, and is technically a `subprocess.run` call according to the [pydocs](https://docs.python.org/3/library/subprocess.html#subprocess.check_output). This is so the exceptions could contain messages from `stderr`.

.github/workflows/tox.yaml
berkayoz: Thank you, will add this and the fix for the linting errors in a new commit.

actions/snap-upgrade.py
addyess: no, sorry -- i noticed on line `163` there's an assignment to `channel`, then line 166 there's another assignment to `channel`.  it seems like the first assignment doesn't do anything

berkayoz: Oh yeah, I just now realized that. Thanks, will update that asap.

addyess:
Integration tests Passed locally: https://paste.ubuntu.com/p/2FTqVRHbQt/

addyess (approved):
These tests are successful.  Thanks for the obviously hard work in this PR

actions/restore.py
kwmonroe: As written, when `loop` == `MAX_WAIT`, we'll gobble up the exception and fall down to `member update` later, which will presumably fail since `member list` has failed.

Perhaps log the event as WARNING until our final loop, then log as ERROR and `raise`.

actions/restore.py
```diff
@@ -227,7 +246,14 @@ def dismantle_cluster():
...
             log('Disconnecting {}'.format(name), hookenv.DEBUG)
-            etcdctl.unregister(data['unit_id'], endpoint)
+            for _ in range(5):
+                try:
+                    etcdctl.unregister(data['unit_id'], endpoint)
+                    break
+                except EtcdCtl.CommandFailed:
+                    # Back-off timer to let cluster settle
+                    log("Disconnecting {} failed, retrying...".format(name), "WARNING")
+                    time.sleep(_ + 1)
```
kwmonroe: Couple thoughts on this loop.  First, it shares a similar problem as the loop above -- it doesn't do anything special if we exhaust `range(5)`.  Shouldn't we raise in that case, or is it safe to continue rendering a conf even though `unregister` has failed? If it *is* safe, i think we should at least log in the last iteration that "disconnect failed, rendering anyway" or something to that effect.

Second, it's a little odd to have 2 loop styles so close together.  Either use a counter var (like the while loop above) or a range (like the for loop here) and be consistent with the sleeps (constant sleep or backoff, pick one).  This isn't a blocker, btw, just a note that consistent logic will be easier on future maintainers.

kwmonroe, line 256: Couple thoughts on this loop.  First, it shares a similar problem as the loop above -- it doesn't do anything special if we exhaust `range(5)`.  Shouldn't we raise in that case, or is it safe to continue rendering a conf even though `unregister` has failed? If it *is* safe, i think we should at least log in the last iteration that "disconnect failed, rendering anyway" or something to that effect.

Second, it's a little odd to have 2 loop styles so close together.  Either use a counter var (like the while loop above) or a range (like the for loop here) and be consistent with the sleeps (constant sleep or backoff, pick one).  This isn't a blocker, btw, just a note that consistent logic will be easier on future maintainers.

lib/etcdctl.py
kwmonroe: Unfortunately, `capture_output` was added in py37, and we need to support bionic (py36).  If we need `stderr`, let's revert back to `check_output` and add a `stderr=subprocess.STDOUT` param and drop the subsequent `log(e.stderr, ERROR)` below.

There are a couple other ways to ensure stdout/stderr; if you prefer logging the streams separately, we can do that with sp_run or Popen. But given the previous impl, it's not clear that merging these with the `stderr=` arg will be any *worse* that what we currently do.

reactive/etcd.py
```diff
@@ -663,15 +663,21 @@ def force_rejoin_requested():
...
     members = etcdctl.member_list()
     # Self Unregistration
-    etcdctl.unregister(members[unit_name]["unit_id"], leader_address)
+    for _ in range(5):
+        try:
+            etcdctl.unregister(members[unit_name]["unit_id"], leader_address)
+            break
+        except EtcdCtl.CommandFailed:
+            # Randomized back-off timer to let cluster settle
+            time.sleep(random.randint(1, 3))
```
kwmonroe: As with previous loop comments, shouldn't we log loop errors here?  If we truly don't care about the `unregister` return (because we're in a stop hook), why loop?  If we do, let's log what we're retrying with a final message and/or raise if we reach the end of our loop.

kwmonroe, line 680: As with previous loop comments, shouldn't we log loop errors here?  If we truly don't care about the `unregister` return (because we're in a stop hook), why loop?  If we do, let's log what we're retrying with a final message and/or raise if we reach the end of our loop.

tests/unit/test_etcdctl.py
```diff
@@ -103,15 +103,15 @@ def test_etcdctl_command(self):
     def test_etcdctl_environment_with_version_2(self, etcdctl):
         """Validate that environment gets set correctly
         spoiler alert; it shouldn't be set when passing --version"""
-        with patch("etcdctl.check_output") as comock:
+        with patch("etcdctl.sp_run") as comock:
```
kwmonroe: If you revert back to `check_output` as per previous comment, you'll need to revert here as well.

kwmonroe, line 106: If you revert back to `check_output` as per previous comment, you'll need to revert here as well.

tests/unit/test_etcdctl.py
```diff
@@ -103,15 +103,15 @@ def test_etcdctl_command(self):
...
             etcdctl.run("member list", api=2)
             api_version = comock.call_args[1].get("env").get("ETCDCTL_API")
             assert api_version == "2"
 
     def test_etcdctl_environment_with_version_3(self, etcdctl):
         """Validate that environment gets set correctly
         spoiler alert; it shouldn't be set when passing --version"""
-        with patch("etcdctl.check_output") as comock:
+        with patch("etcdctl.sp_run") as comock:
```
kwmonroe: If you revert back to `check_output` as per previous comment, you'll need to revert here as well.

kwmonroe, line 114: If you revert back to `check_output` as per previous comment, you'll need to revert here as well.

kwmonroe (changes requested):
See inline comments about the looping logic used here.  The real blocker here is that we can't rely on `sp_run(capture_output)` as that is only valid in py37+, so we'll at least need to refactor that.

actions/restore.py
```diff
@@ -246,14 +252,23 @@ def dismantle_cluster():
...
-            for _ in range(5):
+            loop = 0
+            MAX_WAIT = 10
+
+            while loop < MAX_WAIT:
                 try:
                     etcdctl.unregister(data['unit_id'], endpoint)
                     break
-                except EtcdCtl.CommandFailed:
+                except EtcdCtl.CommandFailed as ex:
```
kwmonroe, line 262:
```suggestion
                except EtcdCtl.CommandFailed as ex:
                    loop = loop + 1
```

reactive/etcd.py
```diff
@@ -671,13 +671,22 @@ def perform_self_unregistration(cluster=None):
...
     # Self Unregistration
-    for _ in range(5):
+    loop = 0
+    MAX_WAIT = 10
+    while loop < MAX_WAIT:
         try:
             etcdctl.unregister(members[unit_name]["unit_id"], leader_address)
             break
-        except EtcdCtl.CommandFailed:
+        except EtcdCtl.CommandFailed as ex:
```
kwmonroe, line 680:
```suggestion
        except EtcdCtl.CommandFailed as ex:
            loop = loop + 1
```

kwmonroe (commented):
@berkayoz thanks for the push for consistency!  i have a couple suggestions below to increment the loop counters (else we could be in those `while` loops forever).  otherwise lgtm!

reactive/etcd.py
berkayoz: Hey, seems like I totally forgot to increment the variable. Thanks for pointing it out and the reviews!

kwmonroe (approved):
LGTM, thanks!

FYI, the integration test was expected to fail here because secrets won't get passed from a fork's PR.

PR #206, opened 2023-01-25
Remove ubuntu user access to etcd tls cert files

samuelallan72:
Remove ubuntu user access to etcd tls cert files

This is for general security hardening.
It means the ubuntu user will no longer have access
to the sensitive tls cert, key, and ca files for etcd
(ie. the files in /var/snap/etcd/common/ ).

An implication of this is that the `etcdctl` command
will no longer be configured or usable for the ubuntu user.
In cases where users previously accessed the shell
on an etcd unit to run etcdctl commands as the ubuntu user,
they will now need to run as root from an interactive shell
(interactive because the config is set in .bash_aliases).

This also means that for any cases where users manually copied
(via ssh or scripts) the cert files from an etcd unit,
they will now need to use sudo.

Interestingly, etcdctl from the local etcd unit with tls
has been broken anyway,
due to 127.0.0.1 not being included in the etcd certs.
See https://github.com/charmed-kubernetes/layer-etcd/pull/204

Closes-Bug: https://bugs.launchpad.net/charm-etcd/+bug/1997531

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -570,15 +570,12 @@ def set_snapd_timer():
 def tls_state_control():
     """This state represents all the complexity of handling the TLS certs.
     instead of stacking decorators, this state condenses it into a single
-    state we can gate on before progressing with secure setup. Also handles
-    ensuring users of the system can access the TLS certificates"""
+    state we can gate on before progressing with secure setup."""
 
     bag = EtcdDatabag()
     if not os.path.isdir(bag.etcd_conf_dir):
         hookenv.log("Waiting for etcd conf creation.")
         return
-    cmd = ["chown", "-R", "root:ubuntu", bag.etcd_conf_dir]
-    check_call(cmd)
     set_state("etcd.ssl.placed")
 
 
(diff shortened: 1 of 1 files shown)


Cynerva:
I'm definitely on board with removing the ubuntu user's access to `server.key` and `server.crt`. The etcd server process is the only thing that needs or should have access to those.

I'm on the fence about restricting access to the `client.key`, `client.crt`, and `ca.crt`. As you note, these are needed to run etcdctl. By making this change, we will encourage operators to run etcdctl via sudo, giving that process access to far more of the filesystem than it needs access to. Of course, the charm already invokes etcdctl as root, so... maybe this is fine.

I guess my main concern is that it would not be consistent with the Kubernetes charms, where we deliberately do give the ubuntu user admin access to the Kubernetes cluster. We do that to make troubleshooting easier, given that the ubuntu user has passwordless sudo anyway.

I don't know. What do you think?

I'll see if I can get a second opinion from other maintainers too.

kwmonroe:
@Cynerva i agree that `ubuntu` doesn't need read perms on `server.[crt|key]` (it got the 440 [from layer-tls-client](https://github.com/charmed-kubernetes/layer-tls-client/blob/main/reactive/tls_client.py#L208)). I also agree that debugging as `ubuntu` on ck8s units is really useful.

That said, `etcd` feels too powerful to trade unintentional access for easier debugging.  My $0.02 is that we should indeed drop `ubuntu`'s access to etcd certs.

Cynerva:
Fair enough. LGTM

Cynerva (approved):

Cynerva:
And thanks @swalladge for the fix :)

PR #207, opened 2023-03-01
delay sharing details until etcd is registered

addyess:
For etcd: Units are sending cluster connection details before etcd is ready. It should delay sending cluster connection details until after etcd has successfully registered

[LP#2008267](https://bugs.launchpad.net/charm-calico/+bug/2008267)

### DIFF
--- a/reactive/etcd.py
+++ b/reactive/etcd.py
@@ -281,6 +281,7 @@ def rerender_config():
 
 
 @when("cluster.joined")
+@when("etcd.registered")
 def set_db_ingress_address(cluster):
     """Send db ingress address to peers on the cluster relation"""
     address = get_ingress_address("db")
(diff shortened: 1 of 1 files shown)


Cynerva (changes requested):
I think [send_cluster_connection_details](https://github.com/charmed-kubernetes/layer-etcd/blob/4b0f16f119546fa89178b525e480eb2ca77ce3c8/reactive/etcd.py#L294) and [send_single_connection_details](https://github.com/charmed-kubernetes/layer-etcd/blob/4b0f16f119546fa89178b525e480eb2ca77ce3c8/reactive/etcd.py#L322) are going to need to wait for the same condition, too. This PR as it stands will delay etcd units from sending their DB ingress addresses to other etcd units. But it won't delay them from sending a connection string to clients, and that connection string includes the local ingress address even if it hasn't been sent to etcd peers yet (see [here](https://github.com/charmed-kubernetes/layer-etcd/blob/4b0f16f119546fa89178b525e480eb2ca77ce3c8/reactive/etcd.py#L308-L311)).

kwmonroe (approved):

PR #208, opened 2023-03-15
Charm to support ipv6 addresses in urls within templates and relation data

addyess:
Address issues 
* [LP#1913348](https://bugs.launchpad.net/charm-etcd/+bug/1913348)
* [LP#1947034](https://bugs.launchpad.net/charm-etcd/+bug/1947034)

urls with ipv6 addresses cannot be rendered as `https://${address}:${port}` because the address itself contains colons

instead each URL must be rendered as `https://[${address}]:${port}`

this PR creates a new `build_uri` method and uses it everywhere previous urls were being created. 

It takes care to ensure that
* ipv4 addresses remain the same, 
* ipv6 addresses are handled correctly
* anything else is handled the same as a ipv4

### DIFF
--- a/lib/etcd_lib.py
+++ b/lib/etcd_lib.py
@@ -38,9 +38,13 @@ def get_ingress_addresses(endpoint_name):
 
 
 def get_ingress_address(endpoint_name):
-    """Returns an ingress-address belonging to the named endpoint, if
-    available. Falls back to private-address if necessary."""
-    return get_ingress_addresses(endpoint_name)[0]
+    """Returns an ingress-address belonging to the named endpoint.
+
+    * Sort addresses by ipv4 first, then ipv6 next
+    * Falls back to private-address if necessary.
+    """
+    all_addrs = get_ingress_addresses(endpoint_name)
+    return sorted(all_addrs, key=lambda i: ip_address(i).version)[0]
 
 
 def get_bind_address(endpoint_name):
@@ -48,8 +52,10 @@ def get_bind_address(endpoint_name):
     belonging to the named endpoint, if available.
     Falls back to private-address if necessary.
 
-        @param endpoint_name the endpoint from where taking the
-        bind address
+    * Sort addresses by ipv4 first, then ipv6 next
+
+    @param endpoint_name the endpoint from where taking the
+    bind address
     """
     try:
         data = network_get(endpoint_name)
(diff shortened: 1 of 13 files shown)


Cynerva (changes requested):
I think get_connection_string needs the same treatment [here](https://github.com/charmed-kubernetes/layer-etcd/blob/900e72bd2072c2c9d386ca8c64b895b3c6122413/lib/etcdctl.py#L208), to fix the connection string that's used in a variety of etcdctl calls and also sent to peers and clients via relations. Other than that, this all looks good to me.

lib/etcdctl.py
```diff
@@ -205,6 +206,6 @@ def get_connection_string(members, port, protocol="https"):
     port and protocol (defaults to https)"""
     connections = []
     for address in members:
-        connections.append("{}://{}:{}".format(protocol, address, port))
+        connections.append(build_uri(protocol, address, port))
```
addyess, line 209: @Cynerva terrific spot!  thank you

lib/etcd_lib.py
```diff
@@ -38,18 +38,24 @@ def get_ingress_addresses(endpoint_name):
...
-    """Returns an ingress-address belonging to the named endpoint, if
-    available. Falls back to private-address if necessary."""
-    return get_ingress_addresses(endpoint_name)[0]
+    """Returns an ingress-address belonging to the named endpoint.
+
+    * Sort addresses by ipv4 first, then ipv6 next
+    * Falls back to private-address if necessary.
+    """
+    all_addrs = get_ingress_addresses(endpoint_name)
+    return sorted(all_addrs, key=lambda i: ip_address(i).version)[0]
```
addyess, line 47: sort the addresses in such a way that ipv4 pops to the top first:

notice that python `sorted` doesn't alter any other order  (ie -- it doesn't deep sort)

```python
>>> i = [1,4,"a",3,"b",5]
>>> sorted(i, key=lambda k: type(k).__name__)
[1, 4, 3, 5, 'a', 'b']
```

lib/etcd_lib.py
```diff
@@ -80,12 +86,21 @@ def get_bind_address(endpoint_name):
...
+    if interfaces:
+        addresses = sorted(
+            [
+                ip_address(addr["address"])
+                for ifc in interfaces
+                for addr in ifc.get("addresses", [])
+                if "address" in addr
+            ],
+            key=lambda addr: addr.version,
+        )
```
addyess, line 101: sort the addresses from the `bind-addresses` by ipv4 first

Cynerva (approved):
:+1:

PR #210, opened 2023-06-05
Redirect nagios cron output to tee

dnegreira:
If etcd is running inside LXD, directly piping the output to the /var/lib/nagios/etcd-alarm-list.txt file won't be allowed by apparmor.

The aim of this patch is to redirect the output through tee as this will be allowed and alarms will be written normally, we will also supress output to stdout in order not to flood the local mta as the errors should be picked up by nagios.

Fixes: LP#2021950


Cynerva (approved):
LGTM, thanks for the fix :)

PR #212, opened 2023-09-19
Stop listening on 127.0.0.1:4001 with http

addyess:
In order to address a security concern with etcd, we should stop listening on port 4001 (http) and instead only use the more secure `https://127.0.0.1:2379` for charm configuration. If the admin wishes to expose 4001 -- they may do so only on the localhost with a new charm config option `bind_with_insecure_http`. 

This PR merges into a branch where I've upgraded the charm to use juju 3.1 for integration testing

* Addresses [lp#2008652](https://bugs.launchpad.net/charm-etcd/+bug/2008652)
* Allows for a config `bind_with_insecure_http` to allow the user to enable if wanted

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -30,6 +30,12 @@ options:
     description: |
       The service binds to all network interfaces if true. The service binds
       only to the first found bind address of each relation if false
+  bind_with_insecure_http:
+    type: boolean
+    default: false
+    description: |
+      The service binds to localhost:4001 with http rather than https
+      which requires no tls certificates if true.
   tls_cipher_suites:
     type: string
     default: ""
--- a/config.yaml
+++ b/config.yaml
@@ -34,8 +34,8 @@ options:
     type: boolean
     default: false
     description: |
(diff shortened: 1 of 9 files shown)


config.yaml
```diff
@@ -30,6 +30,12 @@ options:
     description: |
       The service binds to all network interfaces if true. The service binds
       only to the first found bind address of each relation if false
+  bind_with_insecure_http:
+    type: boolean
+    default: false
+    description: |
+      The service binds to localhost:4001 with http rather than https
+      which requires no tls certificates if true.
```
kwmonroe:
```suggestion
      The service binds to localhost:4001 with http if true. This exposes an insecure
      endpoint for the service and is not recommended in production environments.
```

kwmonroe (approved):
LGTM with minor suggestion to the config description.

I really like the new option to re-create an insecure endpoint if someone really wants it, though I think we should target this for 1.29 vs a 1.28 maint release.

PR #215, opened 2025-01-28
[LP#2096820] Don't push stderr through stdout when running etcdctl

addyess:
### Overview
Addresses [LP#2096820](https://bugs.launchpad.net/charm-etcd/+bug/2096820)

Simply -- better handling of stdout, not letting it be polluted with stderr when a etcdctl command completes successfully with warnings in stderr.

### DIFF
--- a/lib/etcdctl.py
+++ b/lib/etcdctl.py
@@ -177,7 +177,7 @@ class EtcdCtl:
             command.insert(2, endpoints)
 
         try:
-            proc = run(command, env=env, check=True, capture_output=True)
+            proc = run(command, env=env, check=True, capture_output=True, text=True)
         except CalledProcessError as e:
             log(command, "ERROR")
             log(env, "ERROR")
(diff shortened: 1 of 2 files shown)


lib/etcdctl.py
```diff
@@ -178,14 +177,20 @@ def run(self, arguments, endpoints: Optional[str] = None, api=3):
             command.insert(2, endpoints)
 
         try:
-            return check_output(command, env=env, stderr=STDOUT).decode("utf-8")
+            proc = run(command, env=env, check=True, capture_output=True)
```
mateoflorido: We need to decode both stdout and stderr in the except clause, right?

lib/etcdctl.py
```diff
@@ -178,14 +177,20 @@ def run(self, arguments, endpoints: Optional[str] = None, api=3):
...
         except CalledProcessError as e:
             log(command, "ERROR")
             log(env, "ERROR")
             log(e.stdout, "ERROR")
             log(e.stderr, "ERROR")
             raise EtcdCtl.CommandFailed() from e
 
+        if proc.stderr:
```
mateoflorido:
```suggestion
        if proc.stderr.decode("utf-8").strip():
```

mateoflorido (commented):
Minor nits. Would adding `text=True, encoding="utf-8", errors="replace"` make this code less verbose/repetitive by reducing the number of times we decode the standard output/error?

lib/etcdctl.py
addyess: Done

mateoflorido (approved):
LGTM!

PR #216, opened 2025-03-03
feat: Drop focal support from etcd

addyess:
Applicable spec: KU-2847

### Overview

* Drops support for python 3.8 and focal as a part of the 1.33 release

### Rationale

In May 2025, Focal loses it's 5-year LTS status in favor of jammy and noble.


louiseschmidtgen (approved):
LGTM thanks Adam!

